PR #10, opened 2019-10-16
update dependency info in README

nhartner:


keefertaylor:
Thanks! I figured we needed to cut a new release in Maven (including Neil's bundling changes) before updating the README. 

I'm cutting that release now.

PR #11, opened 2019-10-17
Pull pom.xml changes off 1.0.0 branch

keefertaylor:
A bunch of changes got made on the 1.0.0 branch to accommodate publishing to Maven Central. 

Pull them back to this branch in order to cut a new release with Neil's fix. 

A few notes:
- Documentation updates where required in order to make the JavaDoc plugin happy. 
- AFAICT, the protocol buffer compiler does not generate JavaDoc and the JavaDoc plugin only allows excludes by packages. Because of this, a large bit of code is excluded. I added a TODO and task in Asana to track and fix this.
- Updated CI config to drop `--remote` which leaves the submodule at the pinned state, rather than pulling from head on master.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>io.xpring</groupId>
     <artifactId>xpring4j</artifactId>
-    <version>SNAPSHOT</version>
+    <version>HEAD-SNAPSHOT</version>
     <name>xpring4j</name>
     <url>https://github.com/xpring-eng/xpring4j</url>
     <description>
--- a/pom.xml
+++ b/pom.xml
@@ -8,6 +8,7 @@
     <artifactId>xpring4j</artifactId>
     <version>1.0.0</version>
     <name>xpring4j</name>
+    <url>https://github.com/xpring-eng/xpring4j</url>
     <description>
         Xpring4j provides a Java based SDK for interacting with Xpring Platform (XRP/ILP)
     </description>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -6,7 +6,12 @@
 
     <groupId>io.xpring</groupId>
     <artifactId>xpring4j</artifactId>
-    <version>1.0-SNAPSHOT</version>
+    <version>1.0.0</version>
```
sappenin: The `master` branch should never have a release version in it. Instead, checkout a tag for `1.0.0` and release from there. The `master` branch should _always_ be a `SNAPSHOT`.

src/main/java/io/xpring/javascript/JavaScriptSigner.java
```diff
@@ -29,8 +29,12 @@ public JavaScriptSigner() throws JavaScriptLoaderException {
 
     /**
      * Sign the given transaction with the given wallet.
+     *
+     * @throws JavaScriptLoaderException An exception if the javascript could not be loaded.
```
sappenin: Minor nit, but generally this should come last in the Javadoc. For future, consider using checkstyle here -- see https://github.com/hyperledger/quilt for details.

keefertaylor: Linting is a tech debt cleanup task that I'm tracking in Asana :)  I've blocked a day a week as sheerly tech debt clean up days so this is a thing that will get burned down soon.

keefertaylor: Added throws

PR #13, opened 2019-10-17
Use CodeCov

keefertaylor:


.circleci/config.yml
```diff
@@ -22,4 +22,6 @@ jobs:
             git submodule update
       - protobuf/install
 
-      - run: mvn clean test jacoco:report coveralls:report -DrepoToken=$COVERALLS_TOKEN
+      - run: mvn clean test jacoco:report
```
sappenin: If this build step works and gets coverage reports to codecov.io, then I would say keep it. However, in Quilt we do this instead (though this was implemented more than a year ago by someone not me -- so for what that's worth).

https://github.com/hyperledger/quilt/blob/master/.circleci/config.yml#L70

sappenin: Yeah, doesn't seem configured properly since Codecov doesn't seem to be getting the reports: https://codecov.io/gh/xpring-eng/xpring4j/commits

keefertaylor: This looks to be working to me:
https://codecov.io/gh/xpring-eng/xpring4j/pull/13?src=pr&el=continue

There's no 'base report' because nothing has built 'master' and uploaded. However, you can see on the comment above that there's a report generated for this branch and you can browse it at the above link by clicking on the files tab.

sappenin: Oh nice, my bad. LG then...

PR #14, opened 2019-10-21
Validate X-Addresses

keefertaylor:


src/test/java/io/xpring/UtilsTest.java
```diff
@@ -9,19 +9,23 @@
...
+    }
+
     @Test
     public void testIsValidAddressInvalidAlphabet() {
         assertFalse(Utils.isValidAddress("1EAG1MwmzkG6gRZcYqcRMfC17eMt8TDTit"));
     }
 
     @Test
-    public void testIsvValidAddressInvlalidChecksum() {
+    public void testIvValidAddressInvalidClassicAddressChecksum() {
```
nhartner: I think you were intending to fix a typo but you still have one here: `testIvValid...` instead of `testIsValid...`

PR #18, opened 2019-10-30
Add functions to check for valid X-Addresses and Classic Addresses

keefertaylor:
Test cases taken from the tests in xpring-common-js.


src/main/java/io/xpring/Utils.java
```diff
@@ -67,4 +67,32 @@ public static String byteArrayToHex(byte [] bytes) {
...
+            return javaScriptUtils.isValidXAddress(address);
+        } catch (Exception exception) {
+            throw new RuntimeException(exception);
+        }
+    }
+
+    /**
+     * Check if the given string is a valid classic address on the XRP Ledger.
+     *
+     * @param address: A string to validate
```
sappenin: Minor nit: not `:` necessary in Javadoc.

PR #19, opened 2019-11-05
Add Utils method for encoding and decoding X-Addresses

keefertaylor:
Allow users to encode / decode X-Addresses. 

Tests taken from PR that implemented the underlying functionality in JavaScript: https://github.com/xpring-eng/xpring-common-js/pull/43


pom.xml
```diff
@@ -208,7 +208,7 @@
                     <dependency>
                         <groupId>com.puppycrawl.tools</groupId>
                         <artifactId>checkstyle</artifactId>
-                        <version>8.26</version>
+                        <version>8.23</version>
```
keefertaylor: Note: I'm rolling this back to 8.23 because the xml config errors with 8.26

src/main/java/io/xpring/Utils.java
```diff
@@ -68,6 +68,37 @@ public static String byteArrayToHex(byte [] bytes) {
...
 
+    /**
+     * Encode the given classic address and tag into an X-Address.
+     *
+     * @param classicAddress A classic address to encode.
+     * @param tag An optional tag to encode.
+     * @return A new X-Address if inputs were valid, otherwise undefined.
+     * @see <a href="https://xrpaddress.info/">https://xrpaddress.info/</a>
+     */
+    public static String encodeXAddress(String classicAddress, Long tag) {
```
theotherian: Is there a reason to not accept an instance of `ClassicAddress` as the argument here instead of its decomposed fields?

src/main/java/io/xpring/javascript/JavaScriptUtils.java
```diff
@@ -31,6 +32,41 @@ public boolean isValidAddress(String address) {
...
 
+    /**
+     * Encode the given classic address and tag into an X-Address.
+     *
+     * @param classicAddress A classic address to encode.
+     * @param tag An optional tag to encode.
+     * @return A new X-Address if inputs were valid, otherwise null.
+     * @see <a href="https://xrpaddress.info/">https://xrpaddress.info/</a>
+     */
+    public String encodeXAddress(String classicAddress, Long tag) {
```
theotherian: Is there a reason to not accept an instance of `ClassicAddress` as the argument here instead of its decomposed fields?

src/main/java/io/xpring/Utils.java
```diff
@@ -68,6 +68,37 @@ public static String byteArrayToHex(byte [] bytes) {
...
+    }
+
+    /**
+     * Decode a {@link ClassicAddress} from a given X-Address.
+     *
+     * @param xAddress The xAddress to decode.
+     * @return A {@link ClassicAddress} if the inputs were valid, otherwise null.
+     * @see <a href="https://xrpaddress.info/">https://xrpaddress.info/</a>
+     */
+    public static ClassicAddress decodeXAddress(String xAddress) {
```
theotherian: If `null` isn't a valid argument it may be worth using `Objects.requireNonNull(xAddress)` here.

src/main/java/io/xpring/javascript/JavaScriptUtils.java
```diff
@@ -31,6 +32,41 @@ public boolean isValidAddress(String address) {
...
+    }
+
+    /**
+     * Decode a {@link ClassicAddress} from a given X-Address.
+     *
+     * @param xAddress The xAddress to decode.
+     * @return A {@link ClassicAddress} if the inputs were valid, otherwise null.
+     * @see <a href="https://xrpaddress.info/">https://xrpaddress.info/</a>
+     */
+    public ClassicAddress decodeXAddress(String xAddress) {
```
theotherian: If `null` isn't a valid argument it may be worth using Objects.requireNonNull(xAddress) here.

src/main/java/io/xpring/Utils.java
```diff
@@ -68,6 +68,37 @@ public static String byteArrayToHex(byte [] bytes) {
...
+        try {
+            return javaScriptUtils.decodeXAddress(xAddress);
+        } catch (Exception exception) {
```
theotherian: I don't see any checked exceptions in the signature of the that class. Is there another reason for wrapping this in a `RuntimeException` that I'm not understanding?

src/main/java/io/xpring/ClassicAddress.java
```diff
@@ -0,0 +1,45 @@
+package io.xpring;
+
+import io.xpring.javascript.JavaScriptWallet;
+import io.xpring.javascript.JavaScriptWalletFactory;
+import io.xpring.javascript.JavaScriptWalletGenerationResult;
+
+/**
+ * Represents classic address components on the XRP Ledger.
+ */
+public class ClassicAddress {
```
theotherian: Any thoughts on adding equals/hashcode implementations to this class? I don't know if you've used it a lot or have any desire to, but the Java library Immutables makes this a bit easier and works well for simple immutable types like this.

pom.xml
sappenin: Yeah, I just had to do that this morning on another project. :(

src/main/java/io/xpring/ClassicAddress.java
sappenin: @keefertaylor if you want to hop on a bluejeans I can show you how immutables works - it's pretty easy to grok, and will save you a lot of pain here.

src/main/java/io/xpring/Utils.java
sappenin: @keefertaylor One style paradigm I really like in Java is to treat `null` values as a bug. With proper unit testing, this can save a lot of pain over the long-term. My thinking here was inspired by the Google Guava engineers per this blog post: https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained

src/main/java/io/xpring/Utils.java
keefertaylor: This should never happen. Because of the nature of JavaScript we can't be sure a given function exists before running it (or that it won't throw given some inputs). As a result, there's really not much we can do.

If the JavaScript is malformed, this whole library is basically useless so we just crash. There's no recovery if the core is corrupted. The good news is that we'd never be able to pass the tests needed to deploy the library, so this is kind of just an impossible use case. 

Lastly, `Utils` is acting as a facade. You could imagine that one day we might want to replace this with either a native implementation or something else that would not throw. No need to pollute our public API.

sappenin: I think what @theotherian is saying is that you can probably just get rid of the try/catch if `javaScriptUtils.decodeXAddress(xAddress)` doesn't explicitly declare a checked-exception.

src/main/java/io/xpring/ClassicAddress.java
keefertaylor: Done.

src/main/java/io/xpring/javascript/JavaScriptUtils.java
keefertaylor: Done

src/main/java/io/xpring/Utils.java
keefertaylor: Fixed through the rest of the file as well.

keefertaylor:
PTAL

src/main/java/io/xpring/ClassicAddress.java
```diff
@@ -0,0 +1,24 @@
...
+import io.xpring.javascript.JavaScriptWalletGenerationResult;
+import org.immutables.value.Value;
+
+import java.util.Optional;
+
+/**
+ * Represents classic address components on the XRP Ledger.
+ */
+@Value.Immutable
+public abstract class ClassicAddress {
```
sappenin: For future: using an `abstract class` is just fine, and equivalent to what I'm about to say, but using Java interface's with Immutables requires less boilerplate code. E.g., here's the interface equivalent:

```
/**
 * Represents classic address components on the XRP Ledger.
 */
@Immutable
public interface ClassicAddress {
/**
     * The address component of the classic address.
     */
    String address();

    /**
     * The tag component of the classic address.
     */
    Optional<Long> tag();
}
```

keefertaylor: Oh interesting. Their tutorial uses `abstract class`: 
https://immutables.github.io/getstarted.html

I am happier with `interface` anyway so I'll make this change.

keefertaylor: Oh interesting. Their tutorial uses `abstract class`: 
https://immutables.github.io/getstarted.html

I am happier with `interface` anyway so I'll make this change.

PR #20, opened 2019-11-05
Run Linter as Part of CI

keefertaylor:
Use `mvn verify` rather than just `mvn test` so that additional checks are run.


keefertaylor:
Note this will pass once the dependent PR lands:  https://github.com/xpring-eng/xpring4j/pull/19

xpring4j.iml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
theotherian: Is it necessary to add this file to version control? Generally I would advocate for not checking IDE settings in.

keefertaylor: Ugh sorry, this should be git-ignored after the changes in the dependent PR too. Removed.

PR #23, opened 2019-11-06
Disallow sending to classic addresses

keefertaylor:


src/main/java/io/xpring/XpringKitException.java
```diff
@@ -2,7 +2,15 @@
 
 /** Exceptions which occur when working with XpringKit. */
 public class XpringKitException extends Exception {
-    public XpringKitException(String errorMessage) {
-        super(errorMessage);
+    /** Static exception for when a classic address is passed to an X-Address API. */
```
sappenin: FWIW - I still think you should move to a sub-package of `io.xpring` so your'e not squatting on the package (for future code). E.g., this project should probably be `io.xpring.xpring4j` maybe.

keefertaylor:
Filed https://app.asana.com/0/1143722888069470/1149060063533177/f

PR #35, opened 2019-11-08
Factor out Integration Tests

keefertaylor:
Companion to making `XpringClientTests` hermetic: https://github.com/xpring-eng/xpring4j/pull/34


src/test/java/io/xpring/IntegrationTests.java
```diff
@@ -34,20 +30,6 @@ public void getBalanceTest() {
...
     }
 
-    @Test
-    public void getFeeTest() {
-        BigInteger balance = xpringClient.getCurrentFeeInDrops();
-        assertThat(balance).isGreaterThan(BigInteger.ONE).withFailMessage("Fee should have been positive");
-    }
-
-    @Test
-    public void getAccountInfo() {
```
sappenin: Just to clarify, we don't want getting account info to be part of the public contract for `XpringClient`?

keefertaylor: No. We haven't exposed this in other libraries. We plumb through the account balance. Internally, we sometimes call this so that we can get the sequence number.

PR #57, opened 2019-12-18
Pass through `test` parameter when creating Wallets

keefertaylor:
X-Addresses support a special format for TestNet. Xpring-Common-JS currently ignores this parameter and defaults everything to MainNet.

Add support for this parameter in all initializers and factory methods of the `Wallet` class. Pass through this parameter to the address derivation code. 

Make the parameter optional and default to MainNet. This is because I think:
- Most folks will just want mainnet addresses by default
- The behavior of existing code won't change, so we can release this as a point fix

Tested:
- Unit tests

Note that the unit test case is derived from manually checking values on http://xrpaddress.info.

----

Note this was originally implemented in: https://github.com/xpring-eng/xpring-common-js/pull/105

### DIFF
--- a/src/main/java/io/xpring/xrpl/javascript/JavaScriptWalletFactory.java
+++ b/src/main/java/io/xpring/xrpl/javascript/JavaScriptWalletFactory.java
@@ -7,6 +7,7 @@ import org.graalvm.polyglot.PolyglotException;
 import org.graalvm.polyglot.Value;
 
 import java.security.SecureRandom;
+import java.util.Objects;
 
 public class JavaScriptWalletFactory {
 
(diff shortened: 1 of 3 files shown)


src/main/java/io/xpring/xrpl/javascript/JavaScriptWalletFactory.java
```diff
@@ -38,35 +39,54 @@ public String getDefaultDerivationPath() throws JavaScriptLoaderException {
...
 
-    public JavaScriptWalletGenerationResult generateRandomWallet() {
+    /**
+     * Generate a random Wallet.
+     *
+     * @param isTest Whether the address is for use on a test network.
+     * @return A {WalletGenerationResult} containing the artifacts of the generation process.
+     * @throws XpringKitException If wallet generation fails.
+     */
+    public JavaScriptWalletGenerationResult generateRandomWallet(boolean isTest) {
```
theotherian: Looks like this wouldn't have a backwards compatible equivalent; not sure if that was intentional or not. These change would break binary compatibility.

keefertaylor: The `JavascriptWallet` and `JavaScriptWalletFactory` is hidden behind the facade of `Wallet`. The purpose of the `Wallet` class is to act as a public API and a facade. 

We could make this class backwards compatible, but `Wallet` is already backwards compatible and always produces a default value for this argument. 

----

Aside, if you think it's weird that `JavaScriptWallet` and `JavaScriptWallet` factory are hidden behind a `Wallet` class, I actually agree with you. We'll clean this up at some point in the future - it was a dumb design decision I made to get a V0 out the door.

theotherian: Good enough for me 👍

PR #69, opened 2020-01-18
Generate protocol buffers from rippled

keefertaylor:
Long term, [xpring-common-protocol-buffers](https://github.com/xpring-eng/xpring-common-protocol-buffers) is going to disappear and be replaced with an official and well reviewed implementation by the rippled team. 

The implementation in rippled was done in: https://github.com/ripple/rippled/pull/3159.

This change segregates the protocol buffers so that we can do the transition in several phases. As of now:
- /generated/*.proto : 'new' protocol buffers from rippled. Unused, as of now.
- /generated/legacy/*.proto : The same protocol buffers from xpring-common-protocol-buffers which we have been using. 

When we have fully migrated the library to rippled's protocol buffers we can simply delete the `legacy` folder to delete the migration. 

Analog to: https://github.com/xpring-eng/Xpring-JS/pull/108


.gitmodules
```diff
@@ -4,3 +4,7 @@
 [submodule "xpring-common-js"]
 	path = xpring-common-js
 	url = git@github.com:xpring-eng/xpring-common-js.git
+[submodule "rippled"]
+	path = rippled
+	url = https://github.com/ripple/rippled
+	branch = develop
```
sappenin: Just double-checking that `develop` is the correct branch here? I'm guessing the protos haven't made their way to master yet in rippled, so this is just temporary, yes?

keefertaylor: Right now the plan is to use develop (Confirmed with @cjcobb23). In the future, we may consider using checkpoints, but since things should be backwards compatible it's relatively safe to just pull from head.

PR #75, opened 2020-01-29
Sign new protocol buffers from rippled

keefertaylor:

Implement a Signer which uses the rippled protocol buffers and xpring-common-js. 

Note this test case is the same test case from our swift library which is just using magic bytes which I manually verified are correct. 

Analog in Swift: https://github.com/xpring-eng/XpringKit/pull/81

### Context of Change

#### Status
   - [ ] Create a new `DefaultXpringClient` with tests
   - [ ] Read only calls for new `DefaultXpringClient` 
   - [ ] Write calls fro `DeafultXpringClient` (WIP with this PR)

### Type of Change

## Before / After

N/A - Code isn't yet hooked up.

## Test Plan
 New unit tests pass on CI.


src/main/java/io/xpring/xrpl/Signer.java
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * Sign the given transaction with the given wallet.
+     * @param transaction The transaction to sign.
+     * @param wallet The wallet that will sign the transaction.
+     * @return A `SignedTransaction`.
+     */
+    public static byte[] signTransaction(Transaction transaction, Wallet wallet) {
+        try {
+            JavaScriptSigner javascriptSigner = new JavaScriptSigner();
```
nhartner: Would it make sense to create this once as a `static final` field? It seems like it takes several seconds to construct because of the GraalVM JS loading. That could cause extra slowness if multiple transactions need to be signed.

keefertaylor: Fixed.  thanks for the insight.

src/main/java/io/xpring/xrpl/legacy/javascript/LegacyJavaScriptSigner.java
```diff
@@ -20,9 +20,9 @@
 
     public LegacyJavaScriptSigner() throws JavaScriptLoaderException {
         Context context = JavaScriptLoader.getContext();
+        this.transactionClass = JavaScriptLoader.loadResource("LegacyTransaction", context);
```
hbergren: Was the change in order of loading resources intentional?

keefertaylor: no. I think i had vague goals of alphabetization but I clearly didn't even nail that. 

Reverted.

src/main/java/io/xpring/xrpl/legacy/javascript/LegacyJavaScriptSigner.java
```diff
@@ -22,7 +22,7 @@ public LegacyJavaScriptSigner() throws JavaScriptLoaderException {
         Context context = JavaScriptLoader.getContext();
         this.signerClass = JavaScriptLoader.loadResource("Signer", context);
         this.walletClass = JavaScriptLoader.loadResource("Wallet", context);
-        this.transactionClass = JavaScriptLoader.loadResource("Transaction", context);
+        this.transactionClass = JavaScriptLoader.loadResource("LegacyTransaction", context);
```
hbergren: nit: trailing whitespace.

PR #76, opened 2020-01-29
Skeleton of new `DefaultXpringClient`

keefertaylor:
Create a new `DefaultXpringClient` which returns an `unimplemented` error for all methods. Future PRs will implement these methods piecemeal.

Note that XpringClient and XpringClientDecorator are modified to throw exceptions on all methods. Since these methods are doing async / network requests on a remote service, I think it makes sense to allow these methods to throw. LMK if there's a better pattern to use here. 

Bootstrap a single test case that checks unimplemented is thrown. 

Analog to this PR in:
Swift: https://github.com/xpring-eng/XpringKit/pull/87
JS: https://github.com/xpring-eng/Xpring-JS/pull/114

# Background

# Status
#### Status
   - [x] Create a new `DefaultXpringClient` with tests (this PR)
   - [ ] Read only calls for new `DefaultXpringClient` 


src/main/java/io/xpring/xrpl/DefaultXpringClient.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public String send(
+            final BigInteger amount,
+            final String destinationAddress,
+            final Wallet sourceWallet
+    ) throws XpringKitException {
+        throw XpringKitException.unimplemented;
+    }
+
+    @Override
```
theotherian: If you expect this to be a somewhat common and/or default behavior (throwing an unimplemented exception) you could use Java's `default` modifier on the interface you're implementing and make that `throw` the default implementation. Not strictly necessary, just a tool I wanted to suggest.

PR #77, opened 2020-01-29
Abstraction around raw transaction status

keefertaylor:
`XpringClientDecorator` will share a definition across the legacy protocol buffers (from `xpring-common-protocol-buffers`) and the rippled protocol buffers. 

This PR creates an abstraction that wraps the return type from this method in an interface. This will allow us to create a bridge layer between the old and new protocol buffers during the migration.

Analog in:
JS: https://github.com/xpring-eng/Xpring-JS/pull/115
Swift: https://github.com/xpring-eng/XpringKit/pull/88

# Background

# Status

   - [x] [Create a new `DefaultXpringClient` with tests](https://github.com/xpring-eng/xpring4j/pull/76)
   - [ ] Read only calls for new `DefaultXpringClient` (WIP in this PR)
   - [ ] Write calls fro `DeafultXpringClient`

### DIFF
--- a/src/main/java/io/xpring/xrpl/RawTransactionStatus.java
+++ b/src/main/java/io/xpring/xrpl/RawTransactionStatus.java
@@ -30,7 +30,7 @@ public class RawTransactionStatus {
      * Retrieve the transaction status code.
      */
     public String getTransactionStatusCode() {
-        return this.getTransactionStatusCode();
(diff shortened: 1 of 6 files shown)


keefertaylor:
I have a feeling you all are going to tell me to go use an `Immutables` library or similar here, LMK what you recommend using and I'll wire it up.

PR #78, opened 2020-01-29
Implement Transaction Status

keefertaylor:

Implement `getTransactionStatus` for the  `XpringClient` which uses official rippled protocol buffers. 

The set of tests are the same set of tests `LegacyXpringClient` runs against which guarantees consistent behavior between the two implementations.

Analog in:
- JS: https://github.com/xpring-eng/Xpring-JS/pull/133
- Swift: https://github.com/xpring-eng/XpringKit/pull/94

### Context of Change

*Status*

- [x] [Generate new versions of the protocol buffers](https://github.com/xpring-eng/Xpring-JS/pull/108)
- [x] [Migrate existing code to a `legacy` subdirectory and namespace](https://github.com/xpring-eng/Xpring-JS/pull/109)
   - [x] [Create a new `NetworkClient` with fakes](https://github.com/xpring-eng/Xpring-JS/pull/111)
   - [x] [Create a new `DefaultXpringClient` with tests](https://github.com/xpring-eng/Xpring-JS/pull/114/files)
   - [x] Read only calls for new `DefaultXpringClient` (In progress in this PR)

### Type of Change

## Before / After

Users who use the new protocol buffers will be able to check a transaction's status. 

## Test Plan

Unit tests on CI continue to pass.


src/main/java/io/xpring/xrpl/DefaultXpringClient.java
```diff
@@ -85,7 +87,14 @@ public BigInteger getBalance(final String xrplAccountAddress) throws XpringKitEx
...
     public TransactionStatus getTransactionStatus(String transactionHash) throws XpringKitException {
-        throw XpringKitException.unimplemented;
+        RawTransactionStatus transactionStatus = getRawTransactionStatus(transactionHash);
+
+        // Return PENDING if the transaction is not validated.
+        if (!transactionStatus.getValidated()) {
+            return TransactionStatus.PENDING;
+        }
+
+        return transactionStatus.getTransactionStatusCode().startsWith("tes") ? TransactionStatus.SUCCEEDED : TransactionStatus.FAILED;
```
hbergren: So this is not implementing the `UNKNOWN` for transactions we can't handle like Escrow, correct?

src/main/java/io/xpring/xrpl/DefaultXpringClient.java
```diff
@@ -112,6 +121,14 @@ public int getLatestValidatedLedgerSequence() throws XpringKitException {
 
     @Override
     public RawTransactionStatus getRawTransactionStatus(String transactionHash) throws XpringKitException {
-        throw XpringKitException.unimplemented;
+        Objects.requireNonNull(transactionHash);
```
hbergren: non-blocking: Why did you choose to have the `requireNonNull` check in this method rather than `TransactionStatus` above?

src/test/java/io/xpring/xrpl/DefaultXpringClientTest.java
```diff
@@ -116,20 +122,97 @@ public void getBalanceTestWithFailedAccountInfo() throws IOException, XpringKitE
...
         );
 
         // WHEN the balance is retrieved THEN an error is thrown.
         expectedException.expect(Exception.class);
         client.getBalance(XRPL_ADDRESS);
     }
 
+    @Test
+    public void transactionStatusWithUnvalidatedTransactionAndFailureCode() throws IOException, XpringKitException {
+        // GIVEN a XpringClient which will return an invalidated transaction with a failed code.
```
hbergren: nit: `invalidated` => `unvalidated`

src/test/java/io/xpring/xrpl/DefaultXpringClientTest.java
```diff
@@ -116,20 +122,97 @@ public void getBalanceTestWithFailedAccountInfo() throws IOException, XpringKitE
...
+
+        // WHEN the transaction status is retrieved.
+        io.xpring.xrpl.TransactionStatus transactionStatus = client.getTransactionStatus(TRANSACTION_HASH);
+
+        // THEN the status is PENDING.
+        assertThat(transactionStatus).isEqualTo(io.xpring.xrpl.TransactionStatus.PENDING);
+    }
+
+    @Test
+    public void transactionStatusWithValidatedTransactionAndFailureCode() throws IOException, XpringKitException {
```
hbergren: Should we add a test for a transaction that would be `UNKNOWN` but would currently result in returning a `FAILED` code?

That way we're documenting the current behavior, while we have a TODO, and a test to refactor when we come back to implementing that.

src/main/java/io/xpring/xrpl/DefaultXpringClient.java
keefertaylor: This is correct. 

This is tracked in a separate Asana task and we decided this was okay to launch as is (since you can only generate payments here). [Here's the tracking task](https://app.asana.com/0/1143722888069470/1159282373720075).

I plan to follow up on this in the next sprint.

src/main/java/io/xpring/xrpl/DefaultXpringClient.java
keefertaylor: This is an oversight. Added.

src/test/java/io/xpring/xrpl/DefaultXpringClientTest.java
keefertaylor: Agree we should test this. This is kinda a messy and large thing so I split into a separate PR.

I'm more than happy to land this, but the code will have to make assumptions about the old protocol buffers and there's quite a few TODOs. I figured it would be better to haggle on it separately, but I don't mind merging it into this review, if you'd like. 

Here's the code:
https://github.com/xpring-eng/xpring4j/pull/88

keefertaylor:
> I'm happy to haggle over #88 in that PR and not block this stuff on that one test.
I look forward to haggling :)

src/main/java/io/xpring/xrpl/DefaultXpringClient.java
```diff
@@ -85,7 +87,16 @@ public BigInteger getBalance(final String xrplAccountAddress) throws XpringKitEx
      * @return The status of the given transaction.
      */
     public TransactionStatus getTransactionStatus(String transactionHash) throws XpringKitException {
```
intelliot: Per https://xrpl.org/reliable-transaction-submission.html users will typically need to know whether all of the relevant ledgers were checked. (relevant = ones that could have included the transaction, based on the latest validated ledger at the time of submission, and the transaction's LastLedgerSequence)

I don't know whether this is supported in gRPC but we should consider having rippled tell us whether the necessary set of ledgers was searched. The following PR was merged, to be included in rippled 1.5.0 - https://github.com/ripple/rippled/pull/3145

keefertaylor: You've asked this before :) 

It's tracked here: https://app.asana.com/0/1143722888069470/1155647690941929

PR #81, opened 2020-01-29
Add wrapper class for `GetTxResponse`

keefertaylor:

Implement a wrapper class for `GetTxResponse` which conforms to the `RawTransactionStatus` interface.

Analog in JS: https://github.com/xpring-eng/Xpring-JS/pull/132
Analog in Swift: https://github.com/xpring-eng/XpringKit/pull/91

### Context of Change

`XpringClient` uses a decorator pattern where each implementation must conform to an interface. The interface required for the `getRawTransactionStatus` method requires it return a `RawTransactionStatus`. 

The new protocol buffers from rippled return a `GetTxResponse` protocol buffer which does not conform to this interface. This wrapper class conforms a`GetTxResponse` to this interface. This unblocks implementation of the `getRawTransactionStatus` method in rippled. 

*Status*

#### Status
   - [ ] Read only calls for new `DefaultXpringClient`  (this PR)

### Type of Change

## Before / After

No change. This new class is not exposed to the world or hooked up yet. 

## Test Plan

CI Continues to work.


src/main/java/io/xpring/xrpl/RawTransactionStatus.java
```diff
@@ -11,14 +12,25 @@
...
         this.transactionStatusCode = transactionStatus.getTransactionStatusCode();
         this.lastLedgerSequence = transactionStatus.getLastLedgerSequence();
     }
 
+    /**
+     * Create a new RawTransactionStatus from a {@link GetTxResponse} protocol buffer.
+     *
+     * @param getTxResponse The {@link GetTxResponse} to encapsulate.
+     */
+    public RawTransactionStatus(GetTxResponse getTxResponse) {
```
hbergren: Is it normal to overload method signatures like this in Java?

I see `public RawTransactionStatus(...)` right above this.

keefertaylor: I think that this is fine. 

The idea here is that this is a bridge class that conforms both of these inputs (from the new and old protocol buffers) to a single type so that `XpringClientDecorator` is fully compatible. 

I'm going to defer to the local java experts.

theotherian: Perfectly reasonable to have two implementations of a constructor (or method) that accept different arguments. Happens quite often in open source libraries.

PR #84, opened 2020-01-31
Refactor methods for send 

keefertaylor:

This change refactors some helper methods in preparation to support the send method. This is a prefactor to the feature being implemented. 

Analog:
JS: https://github.com/xpring-eng/XpringKit/pull/96
Swift: https://github.com/xpring-eng/XpringKit/pull/96

### Context of Change

*Status*
   - [x] Read only calls for new `DefaultXpringClient`  (this PR)
- [x] [Add a boolean parameter to switch between implementations](https://github.com/xpring-eng/xpring4j/pull/80)

### Type of Change

## Before / After

No change
## Test Plan
CI


src/main/java/io/xpring/xrpl/DefaultXpringClient.java
```diff
@@ -131,4 +130,22 @@ public RawTransactionStatus getRawTransactionStatus(String transactionHash) thro
 
         return new RawTransactionStatus(response);
     }
+
+    private XRPDropsAmount getMinimumFee() {
+        return this.getFee().getDrops().getMinimumFee();
```
hbergren: This is kind of a weird API.

`getFee().getDrops().getMinimumFee()`?

What is this coming from?

keefertaylor: `getFee()` is a method (line 138) which returns a `GetFeeResponse` protocol buffer. 

This protocol buffer accesses members through `get<Foo>` so the `getDrops()` returns a message with getMinimumFee().

This chaining is the same as:
```
GetFeeResponse gfr = this.getFee()
Fee drops = gfr.getDrops(); // this member is oddly named, its generated.
return drops.getMinimumFee().
```

keefertaylor:
> The method naming here is really weird, but we can call it out of scope for this PR. (getFee() should be getFeeResponse()???)

Fixed. 

> Are there any tests for the methods you've implemented here? (If not, should there be?)

This is just a prefactor, so no tests here.  The send function uses all of these methods and stubs them out with fake responses for unit tests. That's in #85. 

In #86 I add integration tests, which makes these methods verify themselves against a live node.

PR #85, opened 2020-02-01
Implement send using new protocol buffers

keefertaylor:

Implement `send` for the  `XpringClient` which uses official rippled protocol buffers. 

Add unit and integration tests. The unit tests are taken from the `LegacyDefaultXpringClient` tests, which ensure that the implementation behaves consistently across protocol buffers. 

Note that for compatibility, `tag` in `Classic` address is changed from a `long` to an `int`. This was a mistype as tags are 32 bits in length ([source](https://xrpl.org/source-and-destination-tags.html))

Analog in:
- JS: https://github.com/xpring-eng/Xpring-JS/pull/136
- Swift https://github.com/xpring-eng/XpringKit/pull/99

### Context of Change

*Status*
   - [x] Read only calls for new `DefaultXpringClient`  (many PRs)
   - [x] Write calls fro `DefaultXpringClient` (this PR)
- [x] [Add a boolean parameter to switch between implementations](https://github.com/xpring-eng/xpring4j/pull/80)

### Type of Change

## Before / After

Users who use the new protocol buffers will be able to send XRP.
## Test Plan

Unit tests on CI continue to pass.  Integration test passes.


src/main/java/io/xpring/xrpl/ClassicAddress.java
```diff
@@ -17,7 +17,7 @@
     /**
      * @return The tag component of the classic address.
      */
-    public Optional<Long> tag();
+    public Optional<Integer> tag();
```
Stormtv: Good catch

PR #88, opened 2020-02-05
[RFC] Add test for Transaction Status with Non Payment Transaction

keefertaylor:

We should test the case where a non-payment transaction's status is queried. This is somewhat complicated to do because we don't have the information we need in the old set of protocol buffers.

Here's a draft of what this would look like. 

### Context of Change

Follow on to #78.

### Type of Change

- [x] New Test Case

## Before / After

New test
## Test Plan

N/A

### DIFF
--- a/src/main/java/io/xpring/xrpl/RawTransactionStatus.java
+++ b/src/main/java/io/xpring/xrpl/RawTransactionStatus.java
@@ -1,6 +1,7 @@
 package io.xpring.xrpl;
 
 import io.xpring.proto.TransactionStatus;
+import rpc.v1.TransactionOuterClass.Transaction;
 import rpc.v1.Tx.GetTxResponse;
 
 /** Encapsulates fields of a raw transaction status which is returned by the XRP Ledger. */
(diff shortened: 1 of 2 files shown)


keefertaylor:
Just an RFC; can totally make this better if we'd like to land it. LMK!

hbergren:
When will we "cut over" from old style protobufs to new style protobufs?

I didn't realize we didn't know transaction type in the old style protobuf.

hbergren:
@keefertaylor ?

keefertaylor:
Sorry about this. 

You can use these protocol buffers today (!) but you need rippled 1.5, which is in beta.  We'll make these available when rippled 1.5 is released officially, though we have yet to give though to when the change will happen.

keefertaylor:
Hey @hbergren, we have a solution for this in JS in
https://github.com/xpring-eng/Xpring-JS/pull/193
and 
https://github.com/xpring-eng/Xpring-JS/pull/197

I'm going to close this PR in favor of just porting that code over here. Feel free to poke me if you disagree.

PR #91, opened 2020-02-06
Rename Exception Class

keefertaylor:

Rename exception class to `XpringException`. `XpringKit` is the name of the exception in the [Swift SDK](http://github.com/xpring-eng/xpringkit). 

Also add a CHANGELOG.md to start keeping track of releases.

### Context of Change

@sappenin  and I wrote xpring4j  in a marathon session and we should have caught this in review but we were moving fast and breaking things. 

### Type of Change

## Before / After

N/A - Clients will need to change code but better to change now and stabilize than later. 

## Test Plan

CI


src/main/java/io/xpring/xrpl/ReliableSubmissionXpringClient.java
```diff
@@ -1,7 +1,5 @@
 package io.xpring.xrpl;
 
-import io.xpring.xrpl.XpringClientDecorator;
```
hbergren: Unused import?

src/test/java/io/xpring/xrpl/DefaultXpringClientTest.java
```diff
@@ -13,10 +12,6 @@
 import static org.mockito.AdditionalAnswers.delegatesTo;
 import io.grpc.ManagedChannel;
 import io.grpc.inprocess.InProcessChannelBuilder;
-import io.xpring.xrpl.XpringKitException;
```
dino-rodriguez: Should this import be updated to `import io.xpring.xrpl.XpringException;` ?

PR #92, opened 2020-02-07
Add `grpcURL` parameter

keefertaylor:

Require users to provide a grpcURL when they instantiate a new `XpringClient`. This allows user to bring their own serving infrastructure and increase decentralization of the network and reliance on Xpring / Ripple Labs. 

This is a breaking change, but a minor inconvenience and this release will already contain breaking changes for the Exception classes name. Better to rip the metaphorical bandaid off all at once. 

### Context of Change

Both the [JavaScript](http://github.com/xpring-eng/xpring-js) and [Swift](http://github.com/xpring-eng/xpringkit) SDKs make this a configurable parameter.  This change standardizes this API across packages. 

The hosted service is listed in the README for a quick start for users. 

### Type of Change

## Before / After

N/A

## Test Plan

CI


src/main/java/io/xpring/xrpl/DefaultXpringClient.java
```diff
@@ -22,10 +22,6 @@
  * @see "https://xrpl.org"
  */
 public class DefaultXpringClient implements XpringClientDecorator {
-    // TODO: Use TLS!
-    // TODO(keefertaylor): Make this configurable.
-    public static final String XPRING_GRPC_URL = "3.14.64.116:50051";
```
hbergren: 👍 on removing a hardcoded IP address

src/main/java/io/xpring/xrpl/DefaultXpringClient.java
```diff
@@ -37,11 +33,9 @@
...
      * No-args Constructor.
      */
-    public DefaultXpringClient() {
+    public DefaultXpringClient(String grpcURL) {
         this(ManagedChannelBuilder
-                .forTarget(XPRING_GRPC_URL)
-                // Let's use plaintext communication because we don't have certs
-                // TODO: Use TLS!
+                .forTarget(grpcURL)
                 .usePlaintext()
```
hbergren: Does this TODO that you removed still apply? Or should we always use plaintext communication?

src/main/java/io/xpring/xrpl/XpringClient.java
```diff
@@ -15,20 +15,24 @@
      * Initialize a new client.
      *
      * The client will use the legacy implementation of protocol buffers.
+     *
+     * @param grpcURL The remote URL to use for gRPC calls.
      */
-    public XpringClient() {
-        this(false);
+    public XpringClient(String grpcURL) {
+        this(grpcURL,false);
```
hbergren: nit: 
```suggestion
        this(grpcURL, false);
```

src/main/java/io/xpring/xrpl/legacy/LegacyDefaultXpringClient.java
```diff
@@ -36,11 +33,9 @@
...
      * No-args Constructor.
      */
-    public LegacyDefaultXpringClient() {
+    public LegacyDefaultXpringClient(String grpcURL) {
         this(ManagedChannelBuilder
-            .forTarget(XPRING_TECH_GRPC_URL)
-            // Let's use plaintext communication because we don't have certs
-            // TODO: Use TLS!
+            .forTarget(grpcURL)
             .usePlaintext()
```
hbergren: Ditto the `TODO` comment about plaintext.

src/main/java/io/xpring/xrpl/DefaultXpringClient.java
keefertaylor: We assume, at least until if/when rippled admin methods are implemented, that we're going to defer TLS. 

The reason is that these calls are doing read / writes to a public ledger and so we weren't overly concerned. There's a somewhat nebulous problem about how to get certs into a rippled node as well, which MAYBE websockets have handled, but we haven't dealt with. 

There's an [asana task](https://app.asana.com/0/1143722888069470/1145001263710841) tracking this and this TODO was written on day one when we were like 'sure we can probably write a java sdk in a few hours', so I had kinda assumed it was redundant. 

Happy to revert and take ownership if it bugs you.

hbergren: I think it's fine. I assume we sign transactions locally/offline?

keefertaylor: Yes, that is correct. The on-node signing API is deprecated and isn't particularly secure. 

We use ripple-keypairs to do this signing.

PR #102, opened 2020-02-20
WIP - enable graal compilation on JDK 11+ to speed up tests

nhartner:
change circleci build to use JDK11 image

Maven build times are long

### Context of Change

As of Java 11, there are flags to enable Graal compilation of JS code. This can speed up execution times for JS methods that are repeatedly executed, as is the case during tests. 

### Type of Change

Enables JVMCI flags for Maven surefire test plugin

## Before / After

No changes to production code

## Test Plan

No changes to production code


keefertaylor:
I also assume this is ready for review even though it has WIP in the title.

nhartner:
Closing because the experimental JVM flag causes some tests to fail now. Not worth trying to make that work.

PR #103, opened 2020-02-21
Add Ilp client to SDK

nkramer44:

Addresses https://app.asana.com/0/0/1160803237671513/f

This PR implements a client to interact with ILP infrastructure, which includes account creation, account retrieval, balance checking, and the ability to send money to someone else on the network.

Also includes integration tests as well as some changes to the build process in CircleCI. The build process is now split into two parts in the same workflow:

1. Compile + Unit Tests
2. Run ILP Integration Tests: This has to be done in a separate job, as the tests require test docker containers to start up, which can only be done in a VM on Circle.

### Context of Change

Including an ILP interface within the XpringSDK will make it far easier for developers to hack/develop on Interledger.  Additionally, the JS implementation of the ILP SDK should likely get used for development of the Xpring ILP Wallet.

In terms of design, I essentially followed the existing design of `XpringClient`.  Now, there is `IlpClient`, which contains an `IlpClientDecorator`, which currently is just `DefaultIlpClient`.  This implementation communicates with Hermes via GRPC using generated classes from the Hermes protobuf definitions.

design doc: https://docs.google.com/document/d/1acrYD1YfMDosh72G1b4cSO83jh4TjjzzYXGMQ1e8VCY/edit#

### Type of Change

### ILP Client
Xpring4j will now expose an interface to interact with Xpring's ILP infrastructure (devnet for now).  Developers can initialize an `IlpClient`, and use that client to:
- createAccount (either a generated account or an account with specified settings)
- getAccount
- getBalance
- sendMoney

Hermes has also been added as a submodule to Xpring4j so Xpring4j can compile its protobufs

### CircleCI Build
- Workflow split into two jobs
     1. Build and run unit tests
     2. Run ILP integration tests

## Test Plan
There are two new test classes: `DefaultIlpClientTests` and `IlpIntegrationTests`.
Both will run in CircleCI, and can be run via the cmd line with `mvn clean test`

If you would only like to run the `IlpIntegrationTests`, you can run 
`mvn clean test -DskipTests -PintegrationTests`

## Future Tasks
For future tasks related to PR.
* Improve/pin down auth
* Keep this client aligned with any future features/endpoints in Hermes
* GRPC auth interceptors should probably come from Hermes. Right now they are just copy+pasted


src/main/java/io/xpring/ilp/IlpClient.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+public class IlpClient {
+    private IlpClientDecorator decoratedClient;
+
+    /**
+     * Initialize a new client with the given options.
+     *
+     */
+    public IlpClient() {
+        this.decoratedClient = new DefaultIlpClient();
```
keefertaylor: We should generally let folks specify the URLs they want to hit I think. I had put in a PR into the XpringClient hierarchy to do this.

src/main/java/io/xpring/ilp/IlpClient.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public GetAccountResponse getAccount(String accountId, String bearerToken) throws XpringException {
+        return decoratedClient.getAccount(accountId, bearerToken);
+    }
+
+    /**
+     * Get the balance of the specified account on the connector.
+     *
+     * @param accountId The account ID to get the balance for.
+     * @param bearerToken Authentication bearer token. TODO: Probably change from string to some wrapped JWT class
```
keefertaylor: Assign an owner here

src/main/java/io/xpring/ilp/IlpClient.java
```diff
@@ -0,0 +1,101 @@
...
+     * @return A {@link BigInteger} with the number of drops in this account.
+     * @throws XpringException If the given inputs were invalid.
+     */
+    public GetBalanceResponse getBalance(final String accountId, final String bearerToken) throws XpringException {
+        return decoratedClient.getBalance(accountId, bearerToken);
+    }
+
+    /**
+     * Send a payment from the given accountId to the destinationPaymentPointer payment pointer
+     * @param destinationPaymentPointer : payment pointer of the receiver
```
keefertaylor: nit: one vertical line break between summary and @params

src/main/java/io/xpring/ilp/IlpClient.java
```diff
@@ -0,0 +1,101 @@
...
+     * @param amount : Amount to send
+     * @param accountId : accountId of the sender
+     * @param bearerToken : auth token of the sender
+     * @return
```
keefertaylor: nit: Fill out @return

src/main/java/io/xpring/ilp/IlpClient.java
```diff
@@ -0,0 +1,101 @@
...
+     * @throws XpringException
+     */
+    public SendPaymentResponse sendPayment(final String destinationPaymentPointer,
+                                           final long amount,
```
keefertaylor: Why type this as a long rather than an arbitrary precision `BigInt`?

src/main/java/io/xpring/ilp/IlpClientDecorator.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    CreateAccountResponse createAccount() throws XpringException;
+
+
+    /**
+     * Create an account on the connector.
+     *
+     *
+     * @param createAccountRequest @return A {@link CreateAccountResponse} containing the account settings that were created on the connector
+     * @param bearerToken
```
keefertaylor: Please doc

src/main/java/io/xpring/ilp/IlpClientDecorator.java
```diff
@@ -0,0 +1,71 @@
...
+     * @throws XpringException If the given inputs were invalid or account creation failed.
+     */
+    CreateAccountResponse createAccount(CreateAccountRequest createAccountRequest, Optional<String> bearerToken) throws XpringException;
+
+    /**
+     * Gets account details for the given account ID
+     *
+     * @param accountId
```
keefertaylor: Please doc

src/main/java/io/xpring/ilp/IlpClientDecorator.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    GetBalanceResponse getBalance(final String accountId, final String bearerToken) throws XpringException;
+
+    /**
+     * Send a payment from the given accountId to the destinationPaymentPointer payment pointer
+     * @param destinationPaymentPointer : payment pointer of the receiver
+     * @param amount : Amount to send
+     * @param accountId : accountId of the sender
+     * @param bearerToken : auth token of the sender
+     * @return
```
keefertaylor: please doc

src/main/java/io/xpring/ilp/grpc/GrpcConstants.java
```diff
@@ -0,0 +1,7 @@
+package io.xpring.ilp.grpc;
+
+import io.grpc.Context;
+
+public final class GrpcConstants {
```
keefertaylor: Does this really need to be public? Could it be private or protected?

src/main/java/io/xpring/ilp/grpc/IlpJwtCallCredentials.java
```diff
@@ -0,0 +1,50 @@
...
+import io.grpc.CallCredentials;
+import io.grpc.Metadata;
+import io.grpc.Status;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.Executor;
+
+/**
+ * TODO: Depend on this from hermes?
```
keefertaylor: Assign an owner. Make this actionable:
```
Investigate if this class can be replaced by depending on Hermes and using FooClass. 
```

src/main/java/io/xpring/ilp/grpc/JwtContext.java
```diff
@@ -0,0 +1,13 @@
+package io.xpring.ilp.grpc;
+
+import io.grpc.Context;
+
+/**
+ * TODO: Depend on this from hermes?
```
keefertaylor: Ditto

src/main/resources/client.properties
```diff
@@ -0,0 +1 @@
+ilp.grpc.url=https://localhost:6565
```
keefertaylor: Just for my knowledge, can you expand on what's happening here. 

Generally when I consume libraries, I assume that I can configure them through injecting configuration in the constructor. This, by contrast, is using a static file to load properties off of and assigning a default. 

Is this a Spring construct? A Java construct? Just something that the ILP team likes? 

Should we support both of these things?

src/test/java/io/xpring/ilp/DefaultIlpClientTest.java
```diff
@@ -0,0 +1,331 @@
...
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Represents the result of a gRPC network call for an object of type T or an error.
+ */
+class GRPCResult<T> {
```
keefertaylor: If you like this class, we should just refactor this to some place common so our tests can use it.

src/test/java/io/xpring/ilp/IlpIntegrationTests.java
```diff
@@ -0,0 +1,172 @@
...
+import org.testcontainers.containers.Network;
+
+import java.util.Optional;
+
+public class IlpIntegrationTests {
+
+  /**
+   * Connector fields
+   */
+  // This should allow hermes to talk to the connector without exposing any ports
```
keefertaylor: Can you elaborate on this. 

Also nit: capitalize `Hermes`

src/test/java/io/xpring/ilp/IlpIntegrationTests.java
```diff
@@ -0,0 +1,172 @@
...
+
+  private static final int CONNECTOR_PORT = 8080;
+
+  private static final int HERMES_PORT = 6565;
+
+  /**
+   *  Start up a connector from the nightly docker image
+   */
+  @ClassRule
+  public static GenericContainer connectorNode = new GenericContainer<>("interledger4j/java-ilpv4-connector:0.2.0") // FIXME use nightly
```
keefertaylor: Assign a TODO owner

src/test/java/io/xpring/ilp/IlpIntegrationTests.java
```diff
@@ -0,0 +1,172 @@
...
+    .withExposedPorts(CONNECTOR_PORT)
+    .withNetwork(network)
+    .withNetworkAliases("connector-host"); // Need this so hermes can communicate with the connector container
```
keefertaylor: Same: I'm not sure what this comment means, and please capitalize Hermes.

src/test/java/io/xpring/ilp/util/IlpConstants.java
```diff
@@ -0,0 +1,52 @@
+package io.xpring.ilp.util;
+
+public class IlpConstants {
+
+  public static String DOT = ".";
+  public static String ILP_OVER_HTTP = "ilpOverHttp";
```
keefertaylor: This constant isn't obvious to me. what are we doing with ILP over HTTP? s

src/main/java/io/xpring/ilp/DefaultIlpClient.java
```diff
@@ -0,0 +1,165 @@
...
+    private final IlpOverHttpServiceGrpc.IlpOverHttpServiceBlockingStub ilpOverHttpServiceStub;
+
+    /**
+     * No-args Constructor.
+     */
+    public DefaultIlpClient() {
+        this(ManagedChannelBuilder
+          .forTarget(XPRING_ILP_GRPC_URL)
+          // Let's use plaintext communication because we don't have certs
+          // TODO: Use TLS!
```
Stormtv: We usually put peoples names next to TODOs so there is some quick way to know who is responsible

```java
// TODO(nkramer): Use TLS!
```

src/main/java/io/xpring/ilp/DefaultIlpClient.java
```diff
@@ -0,0 +1,165 @@
...
+            // Make sure that the configuration file exists
+            URL res = DefaultIlpClient.class.getClassLoader().getResource(propertiesFileName);
+            InputStream inputStream = new FileInputStream(res.getFile());
+            // load the properties file
+            properties.load(inputStream);
+
+            XPRING_ILP_GRPC_URL = properties.getProperty("ilp.grpc.url");
+        } catch (Exception e) {
+            logger.info(String.format("Problem loading config file named %s. Using default configuration.", propertiesFileName));
+            XPRING_ILP_GRPC_URL = "http://localhost:6565";
```
Stormtv: Should this be a static value in the IlpConstants?

.circleci/config.yml
```diff
@@ -28,5 +28,50 @@ jobs:
...
+      - run:
+          name: "Build And Run Unit Tests"
+          command: mvn clean install -DskipITs
+
+  ilp_integration_tests:
+    working_directory: ~/repo
+
+    # Need to run in a VM so TestContainers will work
+    machine:
+      image: ubuntu-1604:201903-01
```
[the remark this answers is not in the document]
nkramer44: Yes, it would.  One solution would be to just run the entire build in a VM, but it slows down the build, so I thought it would be better to build in a docker container and run just the ITs in a VM

src/main/java/io/xpring/ilp/grpc/IlpJwtCallCredentials.java
```diff
@@ -0,0 +1,50 @@
...
+  public void applyRequestMetadata(RequestInfo requestInfo, Executor appExecutor, MetadataApplier applier) {
+    try {
+      applyToken(applier, jwtToken);
+    } catch (RuntimeException e) {
+      applyFailure(applier, e);
+    }
+  }
+
+  @Override
+  public void thisUsesUnstableApi() {}
```
Stormtv: What does this mean? Can we add the explanation to the code?

.circleci/config.yml
```diff
@@ -28,5 +28,50 @@ jobs:
...
+            sudo mv protoc-gen-grpc-web /usr/local/bin/
+            chmod +x /usr/local/bin/protoc-gen-grpc-web
+      # gets the project dependencies and installs sub-module deps
+      - run:
+          name: Install Connector Dependencies
+          command: mvn dependency:go-offline -DskipTests -DskipITs install
+      - run:
+          name: "Run ILP Integration Tests"
+          command: mvn clean verify jacoco:report -DskipTests -PintegrationTests
       - run: bash <(curl -s https://codecov.io/bash)
```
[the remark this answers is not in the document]
nkramer44: Yep you're right

src/main/java/io/xpring/ilp/DefaultIlpClient.java
```diff
@@ -0,0 +1,165 @@
...
+ * A client that can create accounts, get accounts, get balances, and send ILP payments on a connector.
+ */
+public class DefaultIlpClient implements IlpClientDecorator {
+
+    private static String XPRING_ILP_GRPC_URL;
+    private static final Logger logger = LoggerFactory.getLogger(DefaultIlpClient.class);
+
+    // Bring in configuration properties such as GRPC url
+    static {
+        String propertiesFileName = "client.properties";
```
Stormtv: Is this standard practice in Java? Are there ways to do environment variable configurations?

src/test/java/io/xpring/ilp/IlpIntegrationTests.java
Stormtv: `// FIXME (nkramer)`

src/main/java/io/xpring/ilp/DefaultIlpClient.java
```diff
@@ -0,0 +1,165 @@
...
+            properties.load(inputStream);
+
+            XPRING_ILP_GRPC_URL = properties.getProperty("ilp.grpc.url");
+        } catch (Exception e) {
+            logger.info(String.format("Problem loading config file named %s. Using default configuration.", propertiesFileName));
+            XPRING_ILP_GRPC_URL = "http://localhost:6565";
+        }
+    };
+
+    private final BalanceServiceGrpc.BalanceServiceBlockingStub balanceServiceStub;
```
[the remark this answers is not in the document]
nkramer44: This has been a point of confusion for @theotherian and me. It's unclear what GRPC best practices are. I remember in one of our initial meetings about GRPC, you said that there should be a 1 to 1 between protobuf messages and GRPC endpoints (aka CreateAccountRequest, GetAccountRequest even though they return the same fields).  

Putting all of your service methods into one monolithic service seems like an anti-pattern...

Is there somewhere where gRPC best practices are documented?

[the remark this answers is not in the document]
theotherian: @keefertaylor I have a question around this as it's come up in previous PRs around integrating our stuff.

I'm not sure I understand why this is considered a desirable direction/best practice? Is there a best practices document/guide somewhere that outlines this more clearly? It seems like condensing all of your services into a single file would end up, at the very least, resulting in a mess of disparate methods that don't fit in one place.

src/main/resources/client.properties
nkramer44: When I wrote this, I assumed from `XpringClient` that we did not want a library consumer to have to configure the GRPC URL.  Instead of hardcoding a gRPC URL in `DefaultIlpClient`, I thought it would be better to make it configurable via a `.properties` file.

Spring has much more support for this, but in plain Java, this is the equivalent of a `.env` file.  Just more annoying to deal with `.properties` files in Java...

Question is, assuming we will now let the consumer configure the grpc url in the `IlpClient` constructor, do we still want a no-arg constructor?  If so, should we just hard-code a default grpc url, or should we keep this pattern of `.properties` files?

src/main/java/io/xpring/ilp/IlpClient.java
nkramer44: Will do this for the other TODOs, but I'm just going to get rid of this one.  I think we decided that there's no use in wrapping the auth token in some sort of primitive

src/main/java/io/xpring/ilp/DefaultIlpClient.java
sappenin: This is unusual from my perspective, though perhaps it's something that google recommends for bootstrapping grpc?

I would expect to see something like a Factory maybe that can create instances of `DefaultIlpClient` with some sort of properties supplier maybe. It's unclear to me if this is just how XpringSDK works, or if we have some room to make this nicer.

Will discuss verbally with the team today.

src/main/java/io/xpring/ilp/DefaultIlpClient.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    @Override
+    public CreateAccountResponse createAccount() throws XpringException {
+        return this.createAccount(CreateAccountRequest.newBuilder().build(), Optional.empty());
+    }
+
+
+    @Override
+    public CreateAccountResponse createAccount(io.xpring.ilp.CreateAccountRequest createAccountRequest, Optional<String> bearerToken) throws XpringException {
```
sappenin: I want to be sure that we make it clear to outside developers that this method will _probably_ go away (creating accounts is likely going to be something developers do while signed-in to the dev portal, though this is still being debated). 

How about we mark this as `@Experimental` (see Guava) or something so that outsiders don't start relying on this as part of the actual API for XpringSDK+ILP (and also so that we all know this is something we should really think hard about before going live with).

src/main/java/io/xpring/ilp/IlpClient.java
sappenin: ILP only supports unsigned longs per the protocol. We should use `UnsignedLong` here though, if possible. @nkramer44  would that work?

nkramer44: We can use `UnsignedLong` to align with ILP.  I typed it as `long` originally because `SendPaymentRequest` requires a `long`.

src/main/java/io/xpring/ilp/grpc/GrpcConstants.java
nkramer44: We actually don't need this class at all.  It's only used on the server

src/test/java/io/xpring/ilp/IlpIntegrationTests.java
nkramer44: We should actually just use the `0.2.0` image, and make sure to update it whenever the SDK uses a new version of the connector

src/test/java/io/xpring/ilp/util/IlpConstants.java
nkramer44: I'll document this, but auth settings are stored in a `Map` where the keys are in a hierarchical dot notation format.

Instead of writing out keys like "ilpOverHttp.incoming.simple.auth_token" in my tests, I just chose to make them constants (this is what we do in Hermes and on the connector).

nkramer44: Also going to rename this to `IlpAuthConstants` to make it more clear what they're for

src/main/java/io/xpring/ilp/grpc/IlpJwtCallCredentials.java
nkramer44: `CallCredentials` is an abstract class from gRPC with this method, so we need to either implement it or declare it abstract.

src/main/java/io/xpring/ilp/DefaultIlpClient.java
nkramer44: This is going to change anyway

src/main/java/io/xpring/ilp/DefaultIlpClient.java
nkramer44: Does Guava have `@Experimental`? Can't find it anywhere

sappenin: Good call - I think it's actually [@Beta](https://github.com/lowasser/guava-experimental/blob/master/guava/src/com/google/common/annotations/Beta.java)

nkramer44: Going to mark `getAccount` as `@Beta` as well.

src/main/java/io/xpring/ilp/DefaultIlpClient.java
nkramer44: To align this with `DefaultXpringClient` and per @keefertaylor's comment, I'm going to get rid of this config stuff in favor of a constructor with configuration.

src/main/resources/client.properties
nkramer44: @keefertaylor see https://github.com/xpring-eng/xpring4j/pull/103#discussion_r384779408

src/main/java/io/xpring/ilp/DefaultIlpClient.java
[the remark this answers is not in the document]
keefertaylor: Oh, I meant that if you had some message that was like:

```
message Foo {}

message GetFooRequest{}
message GetBarRequest{}
```

then it would be weird to have something like:
```
service BarService {
   rpc foo(getFooRequest) returns (Foo)
   rpc foo(getBarRequest) returns (Foo)
}
```

I can't find a logical explanation of best practice in [buf.build]() or [prototool]() or [Google Cloud's style guide]().

I would look at it as a `service` corresponds a single server API. So I would think that you'd just define `HermesService` with 3 rpcs (get balance, send, make account).

If we ever got to the point where there were two binaries running, say we split the accounts service out into xpring-eng/HermesAccountService and we still have Hermes then I would support two gRPC definitions. 

If you use multiple services, I don't think you can bind them all to the same port right? So you would run balanceService on :5001, accountService on :5002, and sendService on :5003. This is confusing for clients who need to input these URLs. 
In general, I assume that the server handling for gRPC is going to be using resources on a per server basis (if nothing else, open sockets and threads in your binary). 

I am happy to let this go in and I think it's generally a trivial refactor. I'm sorry for being unclear about this.

src/main/java/io/xpring/ilp/DefaultIlpClient.java
```diff
@@ -0,0 +1,139 @@
...
+          .build()
+        );
+    }
+
+    /**
+     * Required-args Constructor, currently for testing.
+     *
+     * @param channel A {@link ManagedChannel}.
+     */
+    public DefaultIlpClient(final ManagedChannel channel) {
```
keefertaylor: I think you should shut down these channels on process termination to avoid zombie threads:
https://github.com/xpring-eng/xpring4j/blob/master/src/main/java/io/xpring/xrpl/DefaultXpringClient.java#L66

src/test/java/io/xpring/GRPCResult.java
```diff
@@ -0,0 +1,36 @@
+package io.xpring;
+
+import java.util.Optional;
+
+/**
+ * Represents the result of a gRPC network call for an object of type T or an error.
```
keefertaylor: Thank you for refactoring this :)

keefertaylor:
+ @amiecorso  as FYI

src/main/java/io/xpring/ilp/DefaultIlpClient.java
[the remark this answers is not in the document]
keefertaylor: Upon looking at this though you've managed to use one RPC port and presumably your integration tests pass so maybe I am confused about how listeners work

src/test/java/io/xpring/ilp/DefaultIlpClientTest.java
```diff
@@ -0,0 +1,303 @@
...
+      .setOriginalAmount(1000)
+      .setAmountDelivered(1000)
+      .setAmountSent(1000)
+      .setSuccessfulPayment(true)
+      .build();
+  }
+
+  @Test
+  public void minimalCreateIlpAccountTest() throws XpringException, IOException {
+    DefaultIlpClient client = getClient();
```
keefertaylor: Could you consider adding GIVEN, WHEN, THEN syntax here and in other tests? 

http://www.draconianoverlord.com/2017/11/28/given-when-then.html

src/main/java/io/xpring/ilp/DefaultIlpClient.java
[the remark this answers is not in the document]
nkramer44: All good, I think pretty much everyone is confused about how listeners work/best practices.  

I'm probably going to refactor the messages a bit, but I'll do that in a later PR.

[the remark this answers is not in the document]
theotherian: > If you use multiple services, I don't think you can bind them all to the same port right? So you would run balanceService on :5001, accountService on :5002, and sendService on :5003. This is confusing for clients who need to input these URLs.
In general, I assume that the server handling for gRPC is going to be using resources on a per server basis (if nothing else, open sockets and threads in your binary).

As far as what we've seen, they bind to the same port without a problem, and I'm not sure why it would cause a problem because the name of the service is part of the URL if I recall correctly. Given that the service is scoped at the resource level, it doesn't seem like you would run into any kind of unpredictable collisions. 

> I would look at it as a service corresponds a single server API. So I would think that you'd just define HermesService with 3 rpcs (get balance, send, make account).

Perhaps this is just a difference in thinking with regard to how a RESTful service would look. For example, in REST you wouldn't put all of your resources under a single path (outside of using a path like `/admin` or `/private` for some kind of permission scoping in the URL), you'd usually have something like `/products`, `/users`, etc. Since it seems like a gRPC service loosely acts as a 1 to 1 with what a RESTful resource would be, in our minds it made sense to separate the two.

On that note, looking at some of the resources you passed along, in this one I would infer that there's at least a loose suggestion that having multiple services is a reasonable thing to do: https://buf.build/docs/lint-checkers#service_suffix

It looks like others have attempted the same in terms of running multiple services on the same port; sounds like it's just a matter of attaching multiple to the same server instance:
https://stackoverflow.com/questions/52634217/access-multiple-grpc-services-over-the-same-connection

nkramer44:
@keefertaylor I made those two changes.  Pending Circle, feel free to merge

PR #106, opened 2020-02-24
Refactors GRPCResult to take actual Throwable upon construction

amiecorso:
The class GRPCResult is used to mock network responses for Xpring client unit tests.  This PR refactors GRPCResult to take an actual instance of a Throwable object upon construction, instead of a message string that is later wrapped in a Throwable.

### Context of Change
A new XpringClient method accountExists contains logic that depends on specific Error responses that may be returned from a gRPC call.  The unit test infrastructure needed to be updated to allow for mocking these specific types of responses.  This infrastructure will also be useful for thorough testing of improved error handling in the SDK.
 

### Type of Change

## Before / After
Before: GRPCResult constructed with a String error message
After: GRPCResult constructed with a pre-cooked Throwable that can be as specific as necessary.

## Test Plan
Unit tests continue to pass.

### DIFF
--- a/src/test/java/io/xpring/xrpl/legacy/LegacyDefaultXpringClientTest.java
+++ b/src/test/java/io/xpring/xrpl/legacy/LegacyDefaultXpringClientTest.java
@@ -26,9 +26,9 @@ import java.util.Optional;
  */
 class GRPCResult<T> {
     private Optional<T> value;
-    private Optional<String> error;
+    private Optional<Throwable> error;
 
-    private GRPCResult(T value, String error) {
+    private GRPCResult(T value, Throwable error) {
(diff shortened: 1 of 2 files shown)


src/test/java/io/xpring/xrpl/legacy/LegacyDefaultXpringClientTest.java
```diff
@@ -26,9 +26,9 @@
  */
 class GRPCResult<T> {
```
keefertaylor: Just noting here that if this code wasn't going away, we should totally de-dupe this class across the tests. I was lazy when I did split the protocol buffers which is why this is this way.

PR #110, opened 2020-02-27
(2/2) Return `Unknown` when transaction status is not a payment transaction 

keefertaylor:

Xpring SDK can only bucket non-partial payment transactions into a TransactionStatus category. The current logic assumes all inputs follow this case. This PR will return an Unknown value if that isn't the case.

Analog in JS:https://github.com/xpring-eng/Xpring-JS/pull/197

### Context of Change

The new rippled protocol buffers give us additional information which lets us make this determination. The legacy protocol buffers will continue the same behavior they always have, since they're more or less in maintenance mode. 

### Type of Change

## Before / After

API behavior of `getTransactionStatus` changes. However, it only changes for clients who have opted into the new protocol buffers, which is (at least for now) undocumented / unannounced.

## Test Plan

New unit tests added.


keefertaylor:
Apologies for the delayed response here, @intelliot:

> But I actually think important use cases require the ability to check the status of any XRPL transaction.

Sure. We're marketing this SDK to users who aren't power users and who just want to make payments. In the future this focus may change if we see significant demand for these features.

Power users who can correctly understand these results are certainly free to bind up their own gRPC network client and throw requests at rippled. They'll get back all of the data they need to interpret however they want. 

> If a transaction is finalized with tesSUCCESS, I consider it to be successful, but it is important for users to check the transaction metadata in order to see what the transaction actually did. As long as the user doesn’t make assumptions about what a transaction did — that is, they look at the transactions outcome — then they are safe.

Our working thesis is that most users aren't going to want to understand advanced details of the ledger and operations; they simply want to move value quickly and safely which is the use case we're optimizing for here. 

> If we are comfortable with having some transactions be considered unknowns, then that behavior should be clearly documented, along with justification for that behavior.

Sure. It sounds like we should document this pretty specifically in our next tech writing refresh. This API would also probably be better named 'getStatusForPayment'. 

I've gone ahead and update README.md and the javadoc for this update. LMK or suggest a specific change if there's something you're looking for.

keefertaylor:
Noted. The attitude we've taken (with regards to what we have in flight) is that we're going to support the most basic of use cases and then work outwards from there based on developer demand.

PR #111, opened 2020-02-28
Upgrade for breaking changes in rippled protocol buffers

keefertaylor:

Update to be compatible with breaking protocol buffer changes in rippled.  Breaking changes were introduced in  https://github.com/ripple/rippled/pull/3254.

Note that protocol buffers are not consumed via submodule. 

### Context of Change

Java protocol buffers have an option called `use_multiple_files` which breaks single `.proto` file into multiple `.java` files.

`Common.proto` has an `Amount` message which maps to `Amount.java`. There's also an `Amount.proto` file, which also maps to `Amount.java`. Running the protocol buffer compiler complains that a conflicting file is being written. 

To get around this, I removed rippled as a submodule and copied in the protocol buffers. I then removed the `use_multiple_files` option from `common.proto`. This gets us building. In the future, we'll fix this upstream in rippled, but their code is frozen for this release already.  Note that this option only changes the java generated code, while the byte code and wire binary format are the same, so we are not breaking compatibility. 

-----

No logic changes are introduced. This change does not touch the public API. 

Changes for protocol buffer compatibility and upgrade xpring-common-js to a version that has compatibility with the new protocol buffers

Changes in the protocol buffers are:
- Change folder / package location of files
- Introduce new layers of abstraction internally.

### Type of Change

## Before / After

Compatible with latest rippled. 

## Test Plan
 CI


keefertaylor:
I replied on the Swift PR. We check in code there because the build system tooling for Swift isn't as good and people don't always have an easy way to install the protocol buffer compiler, so it's easier to just check in the generated artifacts rather than deal with bug reports or complicated set up instructions. 

In this library, we don't check in generated java files. We are checking in the .proto files because we need to make a minor modification to get the code from rippled to build. That fix will land upstream and then we'll go back to using a submodule.

PR #113, opened 2020-03-01
Renames XpringClient to XRPClient

amiecorso:
XRPClient is a more appropriate name for the client class in Xpring SDKs. This PR begins the process of updating this name while maintaining the current implementation of XpringClient for gradual deprecation.

### Context of Change
This PR updates all "XpringClient" related names in the SDK to "XRPClient", such as "DefaultXpringClient" --> "DefaultXRPClient", as well as local variables such as "xpringClient" --> "xrpClient". It also introduces a new top-level client class "XRPClient" while maintaining the original "XpringClient" class for eventual deprecation. The "XpringClient" class now wraps an instance of "DefaultXRPClient".

### Type of Change

## Before / After
Before: Everything named as "XpringClient"
After: top level "XpringClient" still exists, but everything else is now "XRPClient", including identical top-level class by this name.

## Test Plan
CI


amiecorso:
Let me know if:
- there's anything needed to be done here analogous to updating the index.js file such that the new XRPClient class is available as part of the package (maybe that's auto in Java but want to be sure)
- my writeup is good.... feel free to nitpick everything!  I also wasn't sure if 'refactor' was the only appropriate box to check, since technically there's a 'new feature' in the XRPClient... but this is just a rename essentially
- when we're confident this is really ready for review, I'll add another reviewer

keefertaylor:
> there's anything needed to be done here analogous to updating the index.js file such that the new XRPClient class is available as part of the package (maybe that's auto in Java but want to be sure)

Everything that is marked `public` should be accessible outside of the package. 

> my writeup is good.... feel free to nitpick everything! I also wasn't sure if 'refactor' was the only appropriate box to check, since technically there's a 'new feature' in the XRPClient... but this is just a rename essentially

This makes sense to me; we're  just moving and renaming code in non-breaking ways. 

When you write GitHub PR desciptions you can use \` to make things display as `code-formatted` but it's not a big deal. 

> when we're confident this is really ready for review, I'll add another reviewer

I think this looks great so I'm going to add another reviewer since you're OOO.

src/test/java/io/xpring/xrpl/XRPClientIntegrationTests.java
```diff
@@ -0,0 +1,81 @@
...
+        assertThat(transactionStatus).isEqualTo(TransactionStatus.SUCCEEDED);
+    }
+
+    @Test
+    public void sendXRPTest() throws XpringException {
+        Wallet wallet = new Wallet(WALLET_SEED);
+
+        String transactionHash = xrpClient.send(AMOUNT, XRPL_ADDRESS, wallet);
+        assertThat(transactionHash).isNotNull();
+    }
```
hbergren: No Given/When/Then comments?

I guess these tests are small and readable enough

amiecorso: Yeah in this case I was just matching the style for all the other tests which didn't have these, but as the integration tests become more extensive we should probably use.

PR #117, opened 2020-03-04
(1/n) Implement enum for TransactionType

keefertaylor:

Provide an enum for transactions.

Reference implementation: https://github.com/xpring-eng/XpringKit/pull/128

### Context of Change

### Type of Change

## Before / After

No change

## Test Plan

N/A


keefertaylor:
@amiecorso:

> are enums conventionally all-caps constant-style?

Yes:  https://howtodoinjava.com/java/enum/enum-tutorial/#what_is_enum

PR #123, opened 2020-03-09
Align sendPayment API

nkramer44:

Asana task: https://app.asana.com/0/1143722888069470/1165556580466057

- Changed `IlpClient.sendPayment` parameter order and names to better match with other version of the SDK.

### Context of Change

Currently, the public `IlpClient` API is slightly different in parameter name and order across Xpring4j, Xpring-JS, and XpringKit.  After this PR is merged, the other two SDKs should expose the same API for `sendPayment`.  This will make the ILP SDK more consistent across languages and provide a clearer API for users.

### Type of Change

## Before / After

Before, `IlpClient.sendPayment` had the following signature.  
```java
public SendPaymentResponse sendPayment(final String destinationPaymentPointer,
                                       final UnsignedLong amount,
                                       final String accountId,
                                       final String bearerToken) throws XpringException 
```

After, `IlpClient.sendPayment` has the following signature.
```java
public SendPaymentResponse sendPayment(final UnsignedLong amount,
                                       final String destinationPaymentPointer,
                                       final String senderAccountId,
                                       final String bearerToken) throws XpringException 
```

`IlpClientDecorator` and thus `DefaultIlpClient` also have this new signature.

## Test Plan

Because this change is just an API change and not a functional change, the existing test suite covers ILP functionality.

## Future Tasks
- Align Xpring-JS and XpringKit to this signature.  XpringKit will have vanity parameter names, but the actual parameter names will be the same as this.
- Update Xpring-SDK-Demo for java

### DIFF
--- a/src/main/java/io/xpring/ilp/DefaultIlpClient.java
+++ b/src/main/java/io/xpring/ilp/DefaultIlpClient.java
@@ -130,14 +130,14 @@ public class DefaultIlpClient implements IlpClientDecorator {
     }
 
     @Override
-    public SendPaymentResponse sendPayment(String destinationPaymentPointer,
-                                           UnsignedLong amount,
-                                           String accountId,
-                                           String bearerToken) throws XpringException {
+    public SendPaymentResponse sendPayment(final UnsignedLong amount,
+                                           final String destinationPaymentPointer,
+                                           final String senderAccountId,
+                                           final String bearerToken) throws XpringException {
         try {
             SendPaymentRequest request = SendPaymentRequest.newBuilder()
               .setDestinationPaymentPointer(destinationPaymentPointer)
-              .setAccountId(accountId)
+              .setAccountId(senderAccountId)
               .setAmount(amount.longValue())
               .build();
             return ilpOverHttpServiceStub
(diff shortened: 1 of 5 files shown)


src/main/java/io/xpring/ilp/DefaultIlpClient.java
```diff
@@ -130,14 +130,14 @@ public GetBalanceResponse getBalance(final String accountId, final String bearer
     }
 
     @Override
-    public SendPaymentResponse sendPayment(String destinationPaymentPointer,
-                                           UnsignedLong amount,
-                                           String accountId,
-                                           String bearerToken) throws XpringException {
+    public SendPaymentResponse sendPayment(final UnsignedLong amount,
```
sappenin: This is fine for now, but will probably lead to errors. For future, we should consider a value object here so that it's obvious which values are getting set.

nkramer44: Agreed.  @keefertaylor and I decided we should wrap protobuf objects so as not to expose them in the public API.  Once we do this we can consolidate these parameters into one of those wrapped objects.

See https://app.asana.com/0/1143722888069470/1165556580466065

PR #126, opened 2020-03-10
Protobuf wrapper objects (2/3)

nkramer44:
- Flattened `AccountBalanceResponse` and removed `AccountBalance` to better match `GetBalanceResponse`
- Added static `from` method to `AccountBalanceResponse` which converts from `GetBalanceResponse` to `AccountBalanceResponse`
- Changed signatures of `IlpClient` and `IlpClientDecorator` to return `AccountBalanceResponse`
- Updated tests

### Context of Change
See #125 for more detail.
I decided to convert `GetBalanceResponse` to `AccountBalanceResponse` in `DefaultIlpClient` instead of in `IlpClient`.  This way, we can test both the conversion and the rest of the `DefaultIlpClient` logic in the same tests.  **Happy to be told this is wrong :)**

### Type of Change

## Before / After
Before:
`IlpClient` and `IlpClientDecorator` `getBalance` returned a protobuf `GetBalanceResponse`.
After:
`GetBalanceResponse` is converted to `AccountBalanceResponse` in `DefaultIlpClient`, and `IlpClient` and `IlpClientDecorator` implementations return an `AccountBalanceResponse`

## Test Plan
The existing test suite covers the conversion process by comparing the return value of `DefaultIlpClient.getBalance` with a mocked `GetBalanceResponse` that is returned by a fake network client.

Do the same with sendPayment


src/main/java/io/xpring/ilp/DefaultIlpClient.java
```diff
@@ -115,15 +116,18 @@ public GetAccountResponse getAccount(String accountId, String bearerToken) throw
     }
 
     @Override
-    public GetBalanceResponse getBalance(final String accountId, final String bearerToken) throws XpringException {
+    public AccountBalanceResponse getBalance(final String accountId, final String bearerToken) throws XpringException {
```
keefertaylor: This strikes me as an awkward name, why not `AccountBalance`. `Response` seems extra and not conveying useful information.

src/main/java/io/xpring/ilp/model/AccountBalanceResponse.java
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * Constructs an {@link AccountBalanceResponse} from a {@link GetBalanceResponse}
+   *
+   * @param getBalanceResponse a {@link GetBalanceResponse} (protobuf object) whose field values will be used
+   *                           to construct an {@link AccountBalanceResponse}
+   * @return an {@link AccountBalanceResponse} with its fields set via the analogous protobuf fields.
+   */
+  static AccountBalanceResponse from(GetBalanceResponse getBalanceResponse) {
+    return builder()
```
keefertaylor: Note that primitives in protobuf are going to return default values. I.E. if `assetCode` is not set, it'll be set to "". 

As long as you are *sure* all of these fields are always set, that's fine. If you are not sure, then you could make this method safer by:

```
if (!getBalanceResponse.hasAssetCode()) {
  throw malformedResponse
}
assetCode = getBalanceResponse.getAssetCode()
```

I'm leaving it up to you to decide the level of safety you want to have

src/main/java/io/xpring/ilp/DefaultIlpClient.java
nkramer44: I don't feel strongly that it should be called `AccountBalanceResponse`.  I basically copied the Hermes REST classes, and that is what we called it in Hermes. But since this is more public facing, I'm happy to rename to `AccountBalance`

src/main/java/io/xpring/ilp/model/AccountBalanceResponse.java
nkramer44: I'm completely sure that they will always be set. In Hermes, we populate an immutable object (in this case, an `AccountBalanceResponse`) from a network call to the connector, then convert it to a gRCP `GetBalanceResponse`.  All of the fields in this class are required, so none can be set to `null` and therefor `GetBalanceResponse` will not have any default values.

src/test/java/io/xpring/ilp/DefaultIlpClientTest.java
```diff
@@ -52,7 +53,7 @@ public void setUp() throws IOException {
       .setAccountId("bob")
       .setAssetCode("XRP")
       .setAssetScale(9)
-      .setNetBalance(1000)
+      .setNetBalance(110) // clearing + prepaid
```
keefertaylor: small nit: you could refactor to be:
```
int clearingBalance = 100
int prepaidAmount = 10
...
.setNextBalance(clearingBalance + prepaidAmount)
.setClearingBalance(clearingBalance)
.setPrepaidAmount(prepaidAmount)
```

Which is even more clear

PR #127, opened 2020-03-10
Protobuf wrapper objects (3/3)

nkramer44:
- Renamed `SendPaymentRequest` (non-proto) to `PaymentRequest` and `SendPaymentResponse` (non-proto) to `PaymentResponse` to remove any ambiguity between protobuf and non-protobuf objects
- Add `toProto` method for converting a non-protobuf `PaymentRequest` to a protobuf `SendPaymentRequest`
- Add `from` method for converting a protobuf `SendPaymentResponse` to a non-protobuf `PaymentResponse`
- Change `sendPayment` signature to consume a `PaymentRequest` and produce a `PaymentResponse`

### Context of Change
See #125 for more detail.

I decided to convert `PaymentRequest` to `SendPaymentRequest`, as well as to convert `SendPaymentResponse` to `PaymentResponse` in DefaultIlpClient instead of in IlpClient. This way, we can test both the conversion and the rest of the DefaultIlpClient logic in the same tests. **Happy to be told this is wrong :)**

### Type of Change

## Before / After
Before:
- `IlpClient` and `IlpClientDecorator` `sendPayment` consumed multiple parameters to construct a `SendPaymentRequest` in `DefaultIlpClient`
- `IlpClient` and `IlpClientDecorator`  `sendPayment` returned a `SendPaymentResponse` (grpc object)
After:
- `IlpClient` and `IlpClientDecorator`  `sendPayment` consume a `PaymentRequest`, which gets converted to a `SendPaymentRequest` via `PaymentRequest.toProto()` in `DefaultIlpClient`
- `IlpClient` and `IlpClientDecorator`  `sendPayment` return a `PaymentResponse`, which is constructed from a `SendPaymentResponse` which is returned from network calls.

## Test Plan
The existing test suite covers the conversion from a `SendPaymentResponse` to a `PaymentResponse` by comparing the return value of `DefaultIlpClient.sendPayment` with a mocked `SendPaymentResponse` that is returned by a fake network client.


src/main/java/io/xpring/ilp/model/PaymentResponse.java
```diff
@@ -7,10 +7,10 @@
  * A response object containing details about a requested payment
  */
 @Value.Immutable
-public interface SendPaymentResponse {
+public interface PaymentResponse {
```
keefertaylor: I'm fine with whatever name you want to use, but I think `PaymentStatus` or `PaymentResult` is more descriptive here.

nkramer44: I think `PaymentResult` is probably the way to go.  `PaymentStatus` makes it sound like that status can change in the future.

PR #130, opened 2020-03-11
Remove `XpringClient`

keefertaylor:

XpringClient is deprecated in favor of `XRPClient`. Remove this class. 

### Context of Change

We want to re-use the `XpringClient` name space to do integrations which tie together other clients (`XRPClient`, `ILPClient`, `PayIDClient`). This removes the class.

This PR does a global find of 'XpringClient' and replaces things appropriately. As such a few trailing references are cleaned up. 

### Type of Change

## Before / After

Clients should migrate to using `XRPClient` if they are still using `XpringClient`.

## Test Plan

CI


Stormtv:
Nice job going to be so much cleaner :)

PR #131, opened 2020-03-11
Create `getPaymentStatus`, deprecate `getTransactionStatus`

keefertaylor:

This PR is a soft rename of `getTransactionStatus` to `getPaymentStatus`.  

JS Analog: https://github.com/xpring-eng/Xpring-JS/pull/232
Swift Analog: https://github.com/xpring-eng/XpringKit/pull/154

### Context of Change

`getTransactionStatus` can only work with Payments so we should name it appropriately.  

This change deprecates the existing method, migrates all test code and updates the changelog / documentation. 

### Type of Change

## Before / After
Better named methods
## Test Plan
CI


src/main/java/io/xpring/xrpl/XpringClient.java
```diff
@@ -6,8 +6,11 @@
 /**
  * A client that can submit transactions to the XRP Ledger.
  *
+ * @deprecated Please use {@link XRPClient} instead.
+ *
  * @see "https://xrpl.org"
  */
+@Deprecated
```
dino-rodriguez: Should this annotation be outside of the comment?

PR #133, opened 2020-03-11
Remove ILP admin functionality

nkramer44:

Asana task: https://app.asana.com/0/1154633712169520/1165822733813057

- Remove `createAccount` and `getAccount` from `IlpClient`, `IlpClientDecorator`, and `DefaultIlpClient`
- Remove and refactor any tests using those methods

### Context of Change
We shouldn't expose functionality like `createAccount` and `getAccount` in the SDK, even though they will still be exposed on Hermes (for now, this will change in the coming months).  Developers should have to have a Xpring account in order to get an access token/JWT, so will have no reason to create more accounts.

This functionality does not exist in Xpring-JS or XpringKit, so this aligns the APIs as well.

### Type of Change

## Before / After
Before: Can create and get accounts on a connector using Xpring4j
After: Can only `getBalance` and `sendPayment` with Xpring4j

## Test Plan
This is just a reduction in functionality, so no additional tests are needed.  However, all of the integration tests were using `IlpClient.createAccount` to create test accounts on local test containers.  Since this no longer exists, the integration tests now use the gRPC network client directly.


src/main/java/io/xpring/ilp/IlpClient.java
```diff
@@ -1,17 +1,11 @@
...
-import io.xpring.ilp.model.CreateAccountRequest;
 import io.xpring.ilp.model.PaymentRequest;
 import io.xpring.ilp.model.PaymentResult;
 import io.xpring.xrpl.XpringException;
 
 import java.math.BigInteger;
-import java.util.Optional;
 
 /**
  * A client that can create accounts, get accounts, get balances, and send ILP payments on a connector.
```
keefertaylor: Can you update this comment?

Issue #139, opened 2020-03-23
Remove plaintext mode gRPC ILP Client


sappenin:
Need to remove this line https://github.com/xpring-eng/xpring4j/blob/master/src/main/java/io/xpring/ilp/DefaultIlpClient.java#L46

PR #140, opened 2020-03-23
Use TLS

sappenin:
Use TLS


nkramer44:
I'm going to get rid of the test containers in the integration tests. The other SDKs hit a running hermes instance, and it would be nice to see that TLS is working as expected

sappenin:
Fixes #139

PR #142, opened 2020-03-23
(2/9) Convert `Currency` protocol buffer

amiecorso:
Provide a native Java wrapper for the [Currency protobuf](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L41). 

Swift analogue: https://github.com/xpring-eng/XpringKit/pull/126
JS analogue: https://github.com/xpring-eng/Xpring-JS/pull/230

### Context of Change

### Type of Change

## Before / After
No change

## Test Plan
new CI introduced for unit tests.

## Future Tasks

3 - 9 / 9 builds up to Transaction protobuf conversion


amiecorso:
Please see my TODO comment about correctly constructing a ByteString in the FakeXRPProtobufs file

src/main/java/io/xpring/xrpl/model/XRPCurrency.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * @return 3 character currency ASCII code
+     */
+    String name();
+
+    /**
+     * @return 160 bit currency code. 20 bytes
+     */
+    ByteString code();
```
nkramer44: `ByteString` is a google protobuf thing, right? I don't have any experience with them, so might not be a huge help here.  The other option is to make this a `byte[]`, which seems more native java-y to me.  Could be convinced otherwise.

In your `from` method, you could then just do `.code(currency.getCode().toByteArray())`

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -0,0 +1,58 @@
...
+public class FakeXRPProtobufs {
+    // primitive test values
+    static String testCurrencyName = "currencyName";
+    static ByteString testCurrencyCode;
+
+    // TODO(amiecorso): not sure if this is the correct way to construct a ByteString...
+    // trying to create equivalent of new Uint8Array([1, 2, 3]) in JS
+    static {
+        try {
+            testCurrencyCode = ByteString.copyFrom("123", "Utf8");
```
nkramer44: This seems like it should work....

If you were to go with a `byte[]`, you could do `"123".getBytes("UTF8")`, but that probably does the same thing.

src/main/java/io/xpring/xrpl/model/XRPCurrency.java
amiecorso: Love it.  I think making these objects as native-java-y as possible is the goal here, and it doesn't make things much uglier, so I'm going to make that change.

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
amiecorso: the testCurrencyCode still has to be a ByteArray because this is what's actually feeding construction of a real protobuf object, which wants to eat googly things

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -0,0 +1,56 @@
...
+    static ByteString testCurrencyCode;
+
+    static {
+        try {
+            testCurrencyCode = ByteString.copyFrom("123", "Utf8");
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
+        }
+    }
+    /** will use in future fake objects */
```
keefertaylor: Assume you're uncommenting this imminently, otherwise I would remove :)

PR #145, opened 2020-03-24
cleanup javadocs

nkramer44:

Clean up javadocs and imports for ILP classes.

### Context of Change
Javadocs should be accurate and up to date.

### Type of Change

- [x] Documentation Updates

## Before / After
Some incorrect javadocs -> correct javadocs

## Test Plan
NA


src/main/java/io/xpring/ilp/IlpClientDecorator.java
```diff
@@ -5,19 +5,17 @@
...
 public interface IlpClientDecorator {
 
     /**
      * Get the balance of the specified account on the connector.
      *
-     * @param accountId The account ID to get the balance for.
-     * @param bearerToken Authentication bearer token.
-     * @return A {@link BigInteger} with the number of drops in this account.
+     * @param accountId The accountId to get the balance for.
+     * @param bearerToken Access token used for authentication.
```
sappenin: We should clarify that this value should _not_ have the prefix "Bearer ". In fact, we should probably not call this a bearer token, but should call it an `accessToken` so that people aren't confused.

nkramer44: Good point.  Now that I look at how we're handling tokens, xpring4j doesn't check to see if the token starts with "Bearer " before prepending it, whereas the other SDKs do check.  Maybe all 3 SDKs should assume that the token does **not** have a "Bearer " prefix and always prepend the token with "Bearer "

nkramer44: Agreed.  I'll do this in a separate PR.

PR #146, opened 2020-03-24
(3/n) Convert Path/PathElement protocol buffer

amiecorso:
Provide a native Java wrapper for the [Path Element protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L227) and [Path protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L237).

Analogue in Swift SDK: [xpring-eng/XpringKit#127](https://github.com/xpring-eng/XpringKit/pull/127)
Analogue in JS SDK: https://github.com/xpring-eng/Xpring-JS/pull/249

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan

## Future Tasks

4 - 9 / 9 will build up to Transaction proto conversion


amiecorso:
So notice that in the `XRPCurrency` class I do an `.isEmpty()` check before building the object, aborting and returning `null` if one of those fields are empty.  In the Java implementation of these protobuf objects, the default value of these fields are indeed empty (an empty string for `name` and the `ByteString` constant `EMPTY` for `code`).  In the[ Currency protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L41) these are required fields (but of course they can be present with empty values).  For our `XRPCurrency` object... are we ok having an object that has an empty value for its `name` or `code`?  Or should the whole object be `null` in this case?  (One reason for causing the whole object to be `null` given a bad field, is that `null` acts as a flag for higher-order objects to abort their own construction if they contain, for example, a bad (`null`)  `XRPCurrency` object...  Anyway, just trying to get a really consistent schema going before writing all of these things.

src/main/java/io/xpring/xrpl/model/XRPPathElement.java
```diff
@@ -0,0 +1,29 @@
...
+@Value.Immutable
+public interface XRPPathElement {
+    static ImmutableXRPPathElement.Builder builder() {
+        return ImmutableXRPPathElement.builder();
+    }
+
+    String account();
+    XRPCurrency currency();
+    String issuer();
+
```
nkramer44: Can you javadoc this class?

amiecorso: documented.  So I wasn't totally clear on what to write for these fields, so I stole from the information here: https://xrpl.org/paths.html#path-steps.  It's a bit verbose, and we also didn't document most of the JS analogues... so we may need to decide on a standard set of documentation and make this consistent throughout the SDKs.  For the rest of these Java PRs, I'll err on the side of verbose and try to document everything, and then can pare down and standardize from there.

PR #147, opened 2020-03-24
Rename bearerToken to accessToken

nkramer44:

- Rename `bearerToken` to `accessToken` in `IlpClient` methods
- Rename `IlpJwtCallCredentials` to `IlpCredentials`
- Enforce that `accessToken` does not start with `Bearer `

Born out of this conversation: https://github.com/xpring-eng/xpring4j/pull/145#discussion_r397265889

### Context of Change
`IlpClient` methods currently take a `String` called `bearerToken`, which is misleading.  Users should just pass in their access token from the ILP wallet in xpring.io, and not worry about the underlying token scheme being used by the SDK/Hermes.  `accessToken` aligns better with the xpring.io nomenclature and is much clearer.  Additionally, users who pass in an `accessToken` with `Bearer ` prefix will receive a `XpringException` telling them not to. This way, their access token with `Bearer ` won't just work behind the scenes.

### Type of Change

This is not really a breaking change, since users currently get a `XpringException` if they passed in a token with `Bearer `, just with an "Unauthenticated" from Hermes.

## Before / After
- All parameters named `bearerToken` are now named `accessToken`

## Test Plan
Added new test class for `IlpCredentials`
Added negative tests in `DefaultIlpClientTests`

## Future Tasks
Create this same logic/contract in xpring-js and xkit


nkramer44:
This should be manually merged with #145 once #145 is merged with master

src/main/java/io/xpring/ilp/DefaultIlpClient.java
```diff
@@ -67,39 +65,45 @@ public DefaultIlpClient(final ManagedChannel channel) {
...
             GetBalanceResponse response = this.balanceServiceStub
-              .withCallCredentials(IlpJwtCallCredentials.build(bearerToken))
+              .withCallCredentials(IlpCredentials.build(accessToken))
               .getBalance(request);
 
             // Convert protobuf response to AccountBalanceResponse
             return AccountBalance.from(response);
         } catch (StatusRuntimeException e) {
             throw new XpringException(String.format("Unable to get balance for account %s.  %s", accountId, e.getStatus()));
+        } catch (IllegalArgumentException e) {
```
keefertaylor: Don't use one letter variables. Call this `exception` or `illegalArgumentException`.

src/main/java/io/xpring/ilp/DefaultIlpClient.java
```diff
@@ -67,39 +65,45 @@ public DefaultIlpClient(final ManagedChannel channel) {
...
               ilpOverHttpServiceStub
-                .withCallCredentials(IlpJwtCallCredentials.build(bearerToken))
+                .withCallCredentials(IlpCredentials.build(accessToken))
                 .sendMoney(request);
 
             return PaymentResult.from(protoResponse);
 
         } catch (StatusRuntimeException e) {
             throw new XpringException("Unable to send payment. " + e.getStatus());
+        } catch (IllegalArgumentException e) {
```
keefertaylor: ditto

src/main/java/io/xpring/ilp/IlpClient.java
```diff
@@ -27,26 +26,26 @@ public IlpClient(String grpcUrl) {
...
+    public AccountBalance getBalance(final String accountId, final String accessToken) throws XpringException {
+        return decoratedClient.getBalance(accountId, accessToken);
     }
 
     /**
      * Send a payment from the given accountId to the destinationPaymentPointer payment pointer
      *
      * @param paymentRequest a {@link PaymentRequest} with parameters used to send a payment
-     * @param bearerToken : auth token of the sender
+     * @param accessToken : Access token of the sender
```
keefertaylor: There's a random `:` here... copy / paste accident?

src/main/java/io/xpring/ilp/grpc/IlpCredentials.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.concurrent.Executor;
+
+/**
+ * An extension of {@link CallCredentials} which provides a convenient way to
+ * add an Authorization metadata header, and ensures every bearer token
+ * going over the wire is prefixed with 'Bearer '
+ */
+public class IlpCredentials extends CallCredentials {
+
+  public static final String BEARER_SPACE = "Bearer ";
```
keefertaylor: nit: `BEARER_WITH_SPACE` is more verbose but slightly clearer. Otherwise it sounds like "bearer space" is a noun. You could also use `BEARER_PREFIX`

src/main/java/io/xpring/ilp/grpc/IlpCredentials.java
```diff
@@ -0,0 +1,72 @@
...
+      "accessToken cannot start with \"Bearer \""
+    );
+    return new IlpCredentials(accessToken);
+  }
+
+  @Override
+  public void applyRequestMetadata(RequestInfo requestInfo, Executor appExecutor, MetadataApplier applier) {
+    try {
+      applyToken(applier, accessToken);
+    } catch (RuntimeException e) {
```
keefertaylor: ditto on one letter vars

src/main/java/io/xpring/ilp/grpc/IlpCredentials.java
```diff
@@ -0,0 +1,72 @@
...
+   * @param applier A {@link io.grpc.CallCredentials.MetadataApplier}
+   * @param accessToken An access token that does not start with 'Bearer '
+   */
+  protected void applyToken(MetadataApplier applier, String accessToken) {
+    Metadata metadata = new Metadata();
+    metadata.put(Metadata.Key.of("Authorization", Metadata.ASCII_STRING_MARSHALLER), BEARER_SPACE + accessToken);
+    applier.apply(metadata);
+  }
+
+  protected void applyFailure(MetadataApplier applier, Throwable e) {
```
keefertaylor: ditto on one letter vars.

src/main/java/io/xpring/ilp/DefaultIlpClient.java
nkramer44: This is pretty common in Java, but will change anyway

PR #152, opened 2020-03-25
(4/n) Convert IssuedCurrency protocol buffer

amiecorso:
Provide a native Java wrapper for the [IssuedCurrencyAmount protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L28).

Analogue in Swift: [xpring-eng/XpringKit#129](https://github.com/xpring-eng/XpringKit/pull/129)
Analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/250

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan

## Future Tasks

5 - 9 / 9 will build up to Transaction protobuf conversion


src/main/java/io/xpring/xrpl/model/XRPIssuedCurrency.java
```diff
@@ -0,0 +1,46 @@
...
+     * @return Unique account address of the entity issuing the currency.
+     */
+    String issuer();
+
+    static XRPIssuedCurrency from(IssuedCurrencyAmount issuedCurrency) {
+        BigInteger value;
+        try {
+            value = new BigInteger((issuedCurrency.getValue()));
+        } catch (NumberFormatException e) {
+            return null;
```
nkramer44: You might want to just throw the `NumberFormatException` here.  This will likely only happen if there is something wrong on the back end (a bad number is getting passed back), which you will want to know about

src/main/java/io/xpring/xrpl/model/XRPIssuedCurrency.java
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        return builder()
+                .currency(XRPCurrency.from(issuedCurrency.getCurrency()))
+                .value(value)
+                .issuer(issuedCurrency.getIssuer().getAddress())
+                .build();
+    }
+}
```
keefertaylor: nit: new line at end of file

src/main/java/io/xpring/xrpl/model/XRPIssuedCurrency.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * @return The currency used to value the amount.
+     */
+    XRPCurrency currency();
+
+    /**
+     * @return The value of the amount.
+     */
+    BigInteger value();
```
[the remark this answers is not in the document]
amiecorso: This link implies that issued currencies CAN have a negative value: https://xrpl.org/currency-formats.html#issued-currency-precision ... I've also asked Elliot who I think might know based on his RippleAPI involvement.  The other reason for potentially leaving it as a BigInteger is for consistency across the three SDKs

src/main/java/io/xpring/xrpl/model/XRPIssuedCurrency.java
amiecorso: good idea, will do this

PR #160, opened 2020-03-26
(5/n) Convert CurrencyAmount protocol buffer

amiecorso:
Provide a native Java wrapper for the[ CurrencyAmount protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/amount.proto#L10).

Analogue in Swift: [xpring-eng/XpringKit#130](https://github.com/xpring-eng/XpringKit/pull/130)
Analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/252

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan

## Future Tasks

6 - 9 / 9 building up to XRPTransaction protobuf wrapper


keefertaylor:
\- Ted (on vaca), \+ Noah for Java expertise as an alternative :)

src/main/java/io/xpring/xrpl/model/XRPCurrencyAmount.java
```diff
@@ -0,0 +1,56 @@
...
+     * Mutually exclusive fields - only drops XOR issuedCurrency should be set.
+     */
+    @Nullable XRPIssuedCurrency issuedCurrency();
+
+    static XRPCurrencyAmount from(CurrencyAmount currencyAmount) {
+        switch (currencyAmount.getAmountCase()) {
+            // Mutually exclusive: either drops or issuedCurrency is set in an XRPCurrencyAmount
+            case ISSUED_CURRENCY_AMOUNT: {
+                IssuedCurrencyAmount issuedCurrencyAmount = currencyAmount.getIssuedCurrencyAmount();
+                if (issuedCurrencyAmount!= null) {
```
keefertaylor: nit: space before `!=`

keefertaylor: I believe in java you can drop the `!= null` if you want and just write:
```
if (issuedCurrency) {
```

Java evaluated objects that are non-null to be truthy values and objects that are null to be falsy

src/main/java/io/xpring/xrpl/model/XRPCurrencyAmount.java
```diff
@@ -0,0 +1,56 @@
...
+                    XRPIssuedCurrency xrpIssuedCurrency = XRPIssuedCurrency.from(issuedCurrencyAmount);
+                    if (xrpIssuedCurrency != null) {
+                        return builder().issuedCurrency(xrpIssuedCurrency).build();
+                    }
+                }
+                // if AmountCase is ISSUED_CURRENCY_AMOUNT, we must be able to convert this to an XRPIssuedCurrency
+                return null;
+            }
+            case XRP_AMOUNT: {
+                long numeric_drops = currencyAmount.getXrpAmount().getDrops();
```
keefertaylor: Use camel case not snake case for variable names

src/main/java/io/xpring/xrpl/model/XRPCurrencyAmount.java
```diff
@@ -0,0 +1,56 @@
...
+                String drops = Long.toString(numeric_drops);
+                return builder().drops(drops)
+                                .build();
+            }
+            // if no AmountCase is set (e.g. empty CurrencyAmount protobuf was passed to constructor)
+            default:
+                return null;
+        }
+    }
+}
```
keefertaylor: nit: newline at end of file

src/test/java/io/xpring/xrpl/ProtocolBufferConversionTest.java
```diff
@@ -119,4 +116,43 @@ public void convertIssuedCurrencyWithBadValueTest() {
...
+        XRPCurrencyAmount xrpCurrencyAmount = XRPCurrencyAmount.from(FakeXRPProtobufs.issuedCurrencyCurrencyAmount);
+
+        // THEN the result has an issued currency set and no drops amount.
+        assertThat(xrpCurrencyAmount.drops()).isNull();
+        assertThat(xrpCurrencyAmount.issuedCurrency())
+                .isEqualTo(XRPIssuedCurrency.from(FakeXRPProtobufs.issuedCurrencyCurrencyAmount.getIssuedCurrencyAmount()));
+    }
+
+    @Test
+    public void convertCurrencyAmountWithBadInputsTest() {
```
keefertaylor: There's actually another case here where the issued currency input is invalid (which is what we've been testing in other tests). This test case is valid though and we should probably have it. Mind adding a test case for an invalidIssuedCurrency?

Extra credit if you want to add it to JS / Swift, otherwise we can file an Asana issue and reference this comment and we can address at some point in the future.

src/main/java/io/xpring/xrpl/model/XRPCurrencyAmount.java
amiecorso: It actually won't let me...  I get a `Required Type: boolean, Provided Type: XRPIssuedCurrency` error.  Java just likes to hear itself talk.

src/test/java/io/xpring/xrpl/ProtocolBufferConversionTest.java
amiecorso: Great, test case has been added here and can do in JS and Swift as well

amiecorso:
Question for y'all:  notice that in my `XRPCurrencyAmount.from` method, I try to convert the underlying `IssuedCurrency` proto into an `XRPIssuedCurrency`... a process which can now throw a `NumberFormatException`.  Here, I catch that `NumberFormatException` and then return `null` instead of finishing construction of an `XRPCurrencyAmount`.  Do you think that's the behavior we want (which silently handles the error) or do we want to re-throw the exception again?

src/main/java/io/xpring/xrpl/model/XRPCurrencyAmount.java
nkramer44: Yeah that would be nice...

src/main/java/io/xpring/xrpl/model/XRPCurrencyAmount.java
```diff
@@ -0,0 +1,62 @@
...
+    static ImmutableXRPCurrencyAmount.Builder builder() {
+        return ImmutableXRPCurrencyAmount.builder();
+    }
+
+    /**
+     * @return An amount of XRP, specified in drops.
+     * Mutually exclusive fields - only drops XOR issuedCurrency should be set.
+     */
+     @Nullable String drops();
+
```
nkramer44: nit: annotations usually go on their own line in Java.

```java
@Nullable
String drops();
```

src/main/java/io/xpring/xrpl/model/XRPCurrencyAmount.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * @return An amount of an issued currency.
+     * Mutually exclusive fields - only drops XOR issuedCurrency should be set.
+     */
+    @Nullable XRPIssuedCurrency issuedCurrency();
```
nkramer44: same here

src/main/java/io/xpring/xrpl/model/XRPCurrencyAmount.java
```diff
@@ -0,0 +1,62 @@
...
+
+    static XRPCurrencyAmount from(CurrencyAmount currencyAmount) {
```
nkramer44: Can you add a javadoc here?

src/main/java/io/xpring/xrpl/model/XRPCurrencyAmount.java
```diff
@@ -0,0 +1,62 @@
...
+        switch (currencyAmount.getAmountCase()) {
+            // Mutually exclusive: either drops or issuedCurrency is set in an XRPCurrencyAmount
+            case ISSUED_CURRENCY_AMOUNT: {
+                IssuedCurrencyAmount issuedCurrencyAmount = currencyAmount.getIssuedCurrencyAmount();
+                XRPIssuedCurrency xrpIssuedCurrency;
+                if (issuedCurrencyAmount != null) {
+                    try {
+                        xrpIssuedCurrency = XRPIssuedCurrency.from(issuedCurrencyAmount);
+                    } catch (NumberFormatException error) {
```
nkramer44: You shouldn't need to rethrow the error here, I don't think.  If `XRPIssuedCurrency.from` throws an exception, this method will throw the exception.

Also, I know @keefertaylor hates 1 letter names, but it's more normal in Java to do `NumberFormatException e`.  If that really bugs you, let's at least change `error` to `exception` (Not here if this is going away, but more for future reference)

amiecorso:
Thanks for the comments Noah! I'll be sure to apply all of these notes to some of the future PRs that you're not a reviewer on, too.  As well as javadoc ALL the `XRPWhatever.from` methods in the same style you've done for ILP wrapper classes.

PR #161, opened 2020-03-27
(6/n) Convert Payment Protocol Buffers

amiecorso:
Provide a native Java wrapper for the [Payment protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L224).
Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/131
Analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/254

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan

## Future Tasks

7 - 9 / 9 will build up to Transaction protobuf conversion


src/main/java/io/xpring/xrpl/model/XRPPayment.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * @return (Optional) Highest amount of source currency this transaction is allowed to cost.
+     */
+    @Nullable XRPCurrencyAmount sendMax();
+
+    static XRPPayment from(Payment payment) {
+        // amount is required
+        XRPCurrencyAmount amount = XRPCurrencyAmount.from(payment.getAmount().getValue());
+        if (amount == null) { return null; }
```
keefertaylor: You can shorten to 

```
if (!amount) { ... } 
```

src/main/java/io/xpring/xrpl/model/XRPPayment.java
```diff
@@ -0,0 +1,117 @@
...
+                .amount(amount)
+                .destination(destination)
+                .destinationTag(destinationTag)
+                .deliverMin(deliverMin)
+                .invoiceID(invoiceID)
+                .paths(paths)
+                .sendMax(sendMax)
+                .build();
+    }
+}
```
keefertaylor: nit: newline

src/main/java/io/xpring/xrpl/model/XRPPayment.java
amiecorso: This is also giving me an issue about requiring a boolean when I try to write it like that...

src/main/java/io/xpring/xrpl/model/XRPPayment.java
```diff
@@ -0,0 +1,131 @@
...
+            destinationTag = payment.getDestinationTag().getValue();
+        } else {
+            destinationTag = null;
+        }
+
+        // If the deliverMin field is set, it must be able to be transformed into an XRPCurrencyAmount.
+        XRPCurrencyAmount deliverMin;
+        if (payment.hasDeliverMin()) {
+            deliverMin = XRPCurrencyAmount.from(payment.getDeliverMin().getValue());
+            if (deliverMin == null) { return null; }
```
nkramer44: nit: Separate the conditional to different lines

src/main/java/io/xpring/xrpl/model/XRPPayment.java
```diff
@@ -0,0 +1,131 @@
...
+                                     .collect(Collectors.toList());
+        if (paths.isEmpty()) {
+            paths = null;
+        }
+
+        // If the sendMax field is set, it must be able to be transformed into an XRPCurrencyAmount.
+        XRPCurrencyAmount sendMax;
+        if (payment.hasSendMax()) {
+            sendMax = XRPCurrencyAmount.from(payment.getSendMax().getValue());
+            if (sendMax == null) { return null; }
```
nkramer44: same here

nkramer44:
Woops, was a little late to the party

PR #162, opened 2020-03-29
(7/9) Convert Memo protocol buffer

amiecorso:
Provide a native Java wrapper for the [Memo protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L80).

Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/136
Analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/255

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan

8 & 9 / building up to Transaction wrapper


amiecorso:
The Circle failure currently due to outdated Testnet txn hash for integration tests.  Will open separate PR to address from master.

src/main/java/io/xpring/xrpl/model/XRPMemo.java
```diff
@@ -0,0 +1,61 @@
...
+        } else {
+            type = memo.getMemoType().getValue().toByteArray();
+        }
+        return XRPMemo.builder()
+                    .data(data)
+                    .format(format)
+                    .type(type)
+                    .build();
+    }
+}
```
keefertaylor: nit: newline

src/main/java/io/xpring/xrpl/model/XRPMemo.java
```diff
@@ -0,0 +1,61 @@
...
+    @Nullable byte[] format();
+
+    /**
+     * @return Hex value representing characters allowed in URLs.
+     * Conventionally, a unique relation (according to RFC 5988) that defines the format of this memo.
+     */
+    @Nullable byte[] type();
+
+    static XRPMemo from(Memo memo) {
+        byte[] data;
```
keefertaylor: I'd build these up one at a time to make it easier to read. Also, the null value is implicitly assigned, though you can make it explicit. 

```
byte[] data; // Or if you want to be super clear: byte[] data= null; 
if (memo.getMemoData().getValue() != ByteString.EMPTY) {
  data = memo.getMemoData().getValue().toByteArray();
}
        // Intentional blank line; let reader know we're done with mucking around with `data`
byte [] format; 
// and so on
```

If you really wanted to be succinct there, you could just use a ternary operator and do the assignment in two statements:
```
ByteString memoData = memo.getMemoData().getValue();
byte[] data = memoData == ByteString.EMPTY ? null : memoData.toByteArray()
        // Intentional blank line; let reader know we're done with mucking around with `data`
ByteString format = ...
```

I tend to prefer the second though some people find ternary operators hard to read.

src/main/java/io/xpring/xrpl/model/XRPMemo.java
```diff
@@ -0,0 +1,61 @@
...
+        byte[] format;
+        byte[] type;
+        if (memo.getMemoData().getValue() == ByteString.EMPTY) {
```
keefertaylor: I'm surprised that `==` WAI. `==` in java should compare object addresses and I'm surprised an arbitrary `EMPTY` ByteString has the same object address as another `EMPTY` byte string. 

We should probably use `.equals()` here:P
```
if (memo.getMemoData.getValue().equals(ByteString.EMPTY) {
```

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -27,6 +27,32 @@
...
     static int testDestinationTag = 2;
+
+    static ByteString memoDataBytes;
+    static ByteString memoFormatBytes;
+    static ByteString memoTypeBytes;
+
+    static {
+        try {
+            memoDataBytes = ByteString.copyFrom("123", "Utf8");
+        } catch (UnsupportedEncodingException e) {
```
keefertaylor: don't use one letter variable names. Saves time now, difficult to read later :) 

(Yes, I know it's only test code but it's good to have good habits)

src/main/java/io/xpring/xrpl/model/XRPMemo.java
amiecorso: wayyyyyy prettier thank you

src/main/java/io/xpring/xrpl/model/XRPMemo.java
amiecorso: I think that the reason this is WAI is that when you construct a default, empty `MemoData` object, its bytestring value is explicitly set to that constant `ByteString.EMPTY` in the constructor. `.equals` sounds good to me and is also easier to read in the ternary operator

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
amiecorso: unlearning bad habits they taught me at school

amiecorso:
CircleCI seems to be crashing for reasons unrelated to tests...

amiecorso:
\+ Noah cause Dino's out

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -16,8 +16,8 @@
     static {
         try {
             testCurrencyCode = ByteString.copyFrom("123", "Utf8");
-        } catch (UnsupportedEncodingException e) {
-            e.printStackTrace();
+        } catch (UnsupportedEncodingException error) {
```
nkramer44: Again, calling this `e` is a very normal thing in java, but if you must use a full word, let's call this `exception`

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -16,8 +16,8 @@
...
+            error.printStackTrace();
```
nkramer44: Let's use a logger here.  Look at `DefaultIlpClient` on how to set up a logger

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -26,6 +26,32 @@
...
+
+    static ByteString memoDataBytes;
+    static ByteString memoFormatBytes;
+    static ByteString memoTypeBytes;
+
+    static {
+        try {
+            memoDataBytes = ByteString.copyFrom("123", "Utf8");
+        } catch (UnsupportedEncodingException error) {
+            error.printStackTrace();
```
nkramer44: logger and `exception` here as well

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -26,6 +26,32 @@
...
+        }
+    }
+    static {
+        try {
+            memoFormatBytes = ByteString.copyFrom("456", "Utf8");
+        } catch (UnsupportedEncodingException error) {
+            error.printStackTrace();
```
nkramer44: same same

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -26,6 +26,32 @@
...
+        }
+    }
+    static {
+        try {
+            memoTypeBytes = ByteString.copyFrom("789", "Utf8");
+        } catch (UnsupportedEncodingException error) {
+            error.printStackTrace();
```
nkramer44: same same

amiecorso:
If you don't mind checking out my logger!  I saw how to set one up but couldn't find an example of it in-use so hopefully what I've got here is ok.

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -16,8 +20,8 @@
     static {
         try {
             testCurrencyCode = ByteString.copyFrom("123", "Utf8");
-        } catch (UnsupportedEncodingException e) {
-            e.printStackTrace();
+        } catch (UnsupportedEncodingException exception) {
+            logger.info("Can't create testCurrencyCode", exception);
```
nkramer44, line 24: This should be `logger.error` or `logger.warn`

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -26,6 +30,32 @@
...
+            memoFormatBytes = ByteString.copyFrom("456", "Utf8");
+        } catch (UnsupportedEncodingException exception) {
+            logger.info("Can't create memoFormatBytes", exception);
+        }
+    }
+    static {
+        try {
+            memoTypeBytes = ByteString.copyFrom("789", "Utf8");
+        } catch (UnsupportedEncodingException exception) {
+            logger.info("Can't create memoTypeBytes", exception);
```
nkramer44, line 56: same here

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -26,6 +30,32 @@
...
+            memoDataBytes = ByteString.copyFrom("123", "Utf8");
+        } catch (UnsupportedEncodingException exception) {
+            logger.info("Can't create memoDataBytes", exception);
+        }
+    }
+    static {
+        try {
+            memoFormatBytes = ByteString.copyFrom("456", "Utf8");
+        } catch (UnsupportedEncodingException exception) {
+            logger.info("Can't create memoFormatBytes", exception);
```
nkramer44, line 49: same here

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -26,6 +30,32 @@
...
+
+    static ByteString memoDataBytes;
+    static ByteString memoFormatBytes;
+    static ByteString memoTypeBytes;
+
+    static {
+        try {
+            memoDataBytes = ByteString.copyFrom("123", "Utf8");
+        } catch (UnsupportedEncodingException exception) {
+            logger.info("Can't create memoDataBytes", exception);
```
nkramer44, line 42: same

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -37,17 +67,17 @@
     static {
         try {
             testPublicKey = ByteString.copyFrom("123", "Utf8");
-        } catch (UnsupportedEncodingException e) {
-            e.printStackTrace();
+        } catch (UnsupportedEncodingException exception) {
+            logger.info("Can't create testPublicKey", exception);
```
nkramer44, line 71: same

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -37,17 +67,17 @@
...
     }
 
     static ByteString testTransactionSignature;
     static {
         try {
             testTransactionSignature = ByteString.copyFrom("456", "Utf8");
-        } catch (UnsupportedEncodingException e) {
-            e.printStackTrace();
+        } catch (UnsupportedEncodingException exception) {
+            logger.info("Can't create testTransactionSignature", exception);
```
nkramer44, line 80: same

amiecorso:
great, all updated to logger.error!

PR #165, opened 2020-03-30
(8/9) Convert Signer Protocol Buffers

amiecorso:
Provide a native Java wrapper for the [Signer protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L90).

Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/137
Analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/256

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan

## Future Tasks

(9/9) Transaction wrapper


src/main/java/io/xpring/xrpl/model/XRPSigner.java
```diff
@@ -0,0 +1,57 @@
...
+    @Nullable byte[] transactionSignature();
+
+    static XRPSigner from(Signer signer) {
+        String account = signer.getAccount().getValue().getAddress();
+        byte[] signingPublicKey;
+        byte[] transactionSignature;
+        if (account.isEmpty()) {
+            account = null;
+        }
+        if (signer.getSigningPublicKey().getValue() == ByteString.EMPTY) {
```
keefertaylor: Same comments as on 7/9 here.

src/main/java/io/xpring/xrpl/model/XRPSigner.java
```diff
@@ -0,0 +1,57 @@
...
+        } else {
+            transactionSignature = signer.getTransactionSignature().getValue().toByteArray();
+        }
+        return XRPSigner.builder()
+                .account(account)
+                .signingPublicKey(signingPublicKey)
+                .transactionSignature(transactionSignature)
+                .build();
+    }
+}
```
keefertaylor: + new line

PR #167, opened 2020-03-31
(9/9) Convert Transaction Protocol Buffers

amiecorso:
Provide a native Java wrapper for the [Transaction protocol buffer](https://github.com/ripple/rippled/blob/develop/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto).

Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/135
Analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/261

### Context of Change

### Type of Change

## Before / After
No change.

## Test Plan


src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * @return an XRPPayment object representing the additional fields present in a PAYMENT transaction.
+     * @see "https://xrpl.org/payment.html#payment-fields"
+     */
+    XRPPayment paymentFields();
+
+    static XRPTransaction from(Transaction transaction) {
+        String account = transaction.getAccount().getValue().getAddress();
+
+        ByteString accountTxnIDByteString = transaction.getAccountTransactionId().getValue();
```
keefertaylor: I'd shy away from using `Txn` because we use `Transaction` everywhere else and I don't think it's a really common abbreviation.

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -0,0 +1,173 @@
...
+        byte[] accountTransactionID = accountTxnIDByteString.equals(ByteString.EMPTY) ?
+                null : accountTxnIDByteString.toByteArray();
+
+        Long fee = transaction.getFee().getDrops();
+
+        Integer flags = transaction.getFlags().getValue();
+
+        Integer lastLedgerSequence = transaction.getLastLedgerSequence().getValue();
+
+        List<XRPMemo> memos = transaction.getMemosList()
```
keefertaylor: Fancy!

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -72,13 +72,22 @@
         }
     }
 
-    /** will use in future fake objects
+    static ByteString testAccountTransactionID;
+    static {
+        try {
+            testAccountTransactionID = ByteString.copyFrom("789", "Utf8");
+        } catch (UnsupportedEncodingException e) {
```
keefertaylor: Name this as `exception`

keefertaylor:
image

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
amiecorso: image

src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
amiecorso: cool I changed all of them

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -0,0 +1,187 @@
...
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * A transaction on the XRP Ledger.
+ *
+ * @see "https://xrpl.org/transaction-formats.html"
+ */
+// TODO(amiecorso): Modify this object to use X-Address format.
```
hbergren, line 18: Just double checking that you wanted to commit this TODO.

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -0,0 +1,187 @@
...
+                .lastLedgerSequence(lastLedgerSequence)
+                .memos(memos)
+                .sequence(sequence)
+                .signers(signers)
+                .signingPublicKey(signingPublicKey)
+                .sourceTag(sourceTag)
+                .transactionSignature(transactionSignature)
+                .type(type)
+                .paymentFields(paymentFields)
+                .build();
```
hbergren, line 185: Nice.

src/test/java/io/xpring/xrpl/ProtocolBufferConversionTest.java
```diff
@@ -289,4 +289,90 @@ public void convertSignerWithAllFieldsSetTest() {
...
+                .stream()
+                .map(signer -> XRPSigner.from(signer))
+                .collect(Collectors.toList()));
+        assertThat(xrpTransaction.signingPublicKey())
+                .isEqualTo(transactionProto.getSigningPublicKey().getValue().toByteArray());
+        assertThat(xrpTransaction.sourceTag()).isEqualTo(transactionProto.getSourceTag().getValue());
+        assertThat(xrpTransaction.transactionSignature())
+                .isEqualTo(transactionProto.getTransactionSignature().getValue().toByteArray());
+        assertThat(xrpTransaction.type()).isEqualTo(TransactionType.PAYMENT);
+        assertThat(xrpTransaction.paymentFields()).isEqualTo(XRPPayment.from(transactionProto.getPayment()));
```
hbergren, line 325: This seems a little odd to me. This looks like you're basically asserting that you wrote this same logic in the implementation, like:
```java
        assertThat(xrpTransaction.memos()).isEqualTo(transactionProto.getMemosList()
                .stream()
                .map(memo -> XRPMemo.from(memo))
                .collect(Collectors.toList()));
```

Is it too hard to have a set of hardcoded data/values to compare against?

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
amiecorso: Yes, there is a consistent TODO across all the SDKs and will address in future PR

src/test/java/io/xpring/xrpl/ProtocolBufferConversionTest.java
amiecorso: This is a good idea - I made myself a TODO among quite a few other SDK cleanups that will be happening after getting these initial wrapper classes / paymentHistory merged.  I think it would make sense to make another test object that can live in the FakeXRPProtobufs alongside the object whose conversion it's representing.

PR #171, opened 2020-03-31
Upgrade Resources on CircleCI

keefertaylor:

Increase the resources available to CI builds.

### Context of Change

@amiecorso reports seeing occasional failures of our tests on CI with [exit code 137](https://support.circleci.com/hc/en-us/articles/115014359648-Exit-code-137-Out-of-memory), which represents an out of memory crash. 

Bump our machine to a `large` executor, rather than the default `medium` executor ([Docs](https://circleci.com/docs/2.0/configuration-reference/#docker-executor)).

This should move us from:
2vCPU -> 4vCPU
4GB RAM -> 8 GB RAM

There's a Medium+ tier between the current and new tier, but these tests run a bit slowly anyway and I'd like to see us get higher performance. Likewise, we could bump for higher performance, but I don't want to be wasteful unnecessarily. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Builds hopefully don't crash. We'll continue to monitor. 

## Test Plan

CI (fingers crossed this run doesn't die)


keefertaylor:
Ping @tedkalaw

PR #172, opened 2020-03-31
Add Interface and Fake for XRPClient

keefertaylor:

Add an interface and a fake for `XRPClient`.

### Context of Change

In the future, a Xpring component will bridge smaller components (ex. PayID, XRP). In order to test interactions with the Xpring component we will need fakes. This PR adds an interface and a fake. 

I'm open to better names that `XRPClientInterface`.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

No change. This is a prefactor for a larger feature.

## Test Plan

CI

### DIFF
--- a/src/main/java/io/xpring/xrpl/XRPClientInterface.java
+++ b/src/main/java/io/xpring/xrpl/XRPClientInterface.java
@@ -1,0 +1,38 @@
+package io.xpring.xrpl;
+
+import io.xpring.xrpl.legacy.LegacyDefaultXRPClient;
+
+import java.math.BigInteger;
+
+public interface XRPClientInterface {
+    /**
+     * Get the balance of the specified account on the XRP Ledger.
+     *
+     * @param xrplAccountAddress The X-Address to retrieve the balance for.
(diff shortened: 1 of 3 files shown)


src/main/java/io/xpring/xrpl/XRPClientInterface.java
```diff
@@ -0,0 +1,38 @@
+package io.xpring.xrpl;
+
+import io.xpring.xrpl.legacy.LegacyDefaultXRPClient;
```
nhartner, line 3: unused import from what I can tell. FYI, IntelliJ has a checkbox on the "Commit Changes" modal to automatically clean up unused imports. It's unchecked by default but I recommend checking it.

image

keefertaylor: Thanks. 

I actually have a TODO to make our linter config actually work (it's committed and it runs but it doesn't auto fix or error, just prints warnings). I'd been mostly deferring these style things until that landed and figured a computer program would just do it for me. 

Cleaned up this one though.

PR #173, opened 2020-04-01
Implements paymentHistory (1/2)

amiecorso:
Implements `paymentHistory` method in `DefaultXRPClient` and related classes.  Does not yet expose this method in `XRPClient` public API.

Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/151
Analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/293

### Context of Change
This PR implements `paymentHistory` and adds additional unit test cases and testing helper infrastructure.

Note this method still does not exist in the public API (will put in separate PR) and thus changelogs etc are not updated and integration tests haven't been written.

### Type of Change

## Before / After
`paymentHistory` and test infrastructure now exists in `DefaultXRPClient`

## Test Plan
CircleCI

## Future Tasks

Expose `paymentHistory` in public API (`XRPClient`), like [this one](https://github.com/xpring-eng/XpringKit/pull/148/files) in Swift SDK


src/main/java/io/xpring/xrpl/ReliableSubmissionXRPClient.java
```diff
@@ -61,6 +64,11 @@ public RawTransactionStatus getRawTransactionStatus(String transactionHash) thro
         return this.decoratedClient.getRawTransactionStatus(transactionHash);
     }
 
+    @Override
+    public List<XRPTransaction> paymentHistory(String address) throws XpringException {
```
keefertaylor, line 68: Just a heads up that #174 renames `XpringException` to `XRPException` in case that lands before this one and you get build errors.

intelliot:
What’s our current thinking about how large history should be handled (pagination)?

amiecorso:
Hi Elliot - I followed up with Keefer about pagination and our current thinking is: we've been striving for an MVP of some of these SDK features, and then can add this change in a non-breaking way.  It is a good point, and I'm going to create an Asana task and put it in the "to be prioritized" section so we can weigh its current relative importance, and then implement it consistently across the three SDKs.

PR #174, opened 2020-04-01
Standardize Exception Format in Xpring4J

keefertaylor:

Create an `Exception` class, with an associated `ExceptionType` enum in all packages (`xrpl`, `ilp`, `payid`) of Xpring4j.

### Context of Change

Priot to this PR we just threw `XpringException`s all over which was a bit of a catch all. This brings standardization to our APIs for developers and allows them to more meaninfully handle errors. 

Additionally, I'll soon introduce a `xpring` package, that will want it's own `XpringException` class, so I need to free up that name space.  

Apologies to reviewers for the long PR; this is all mostly renaming and I can split it up if you'd like. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Throw more specific and useful exceptions.

## Test Plan

CI


src/main/java/io/xpring/ilp/DefaultIlpClient.java
```diff
@@ -78,16 +77,16 @@ public AccountBalance getBalance(final String accountId, final String accessToke
             // Convert protobuf response to AccountBalanceResponse
             return AccountBalance.from(response);
         } catch (StatusRuntimeException statusRuntimeException) {
-            throw new XpringException(String.format("Unable to get balance for account %s.  %s", accountId, statusRuntimeException.getStatus()));
+            throw new IlpException(IlpExceptionType.UNKNOWN, String.format("Unable to get balance for account %s.  %s", accountId, statusRuntimeException.getStatus()));
```
nkramer44, line 80: This is good for now.  I will do something similar to what I am doing in the JS and Swift PRs to translate StatusRuntimeExceptions to IlpExceptions

keefertaylor: Yup. I figured I'd give you a start and then you could just make a breaking change when required.

PR #175, opened 2020-04-02
Remove `getTransactionStatus` method.

keefertaylor:

This method was deprecated in the last release. Remove it in favor of the method named `getPaymentStatus` which is more descriptive.

### Context of Change

Generally we should provide a longer deprecation cycle but the next release is going to:
- Remove `XpringClient`
- Change the default to use rippled's proto implementation

So I think this is a minor cut compared to those two changes and we should just rip the bandage off at once. 

### Type of Change

## Before / After

No deprecated methods. 

## Test Plan

CI


keefertaylor:
\- Austin (on vacation)
\+ New devx-reviewer

PR #176, opened 2020-04-02
Translate gRPC Exceptions (Overrides #174)

nkramer44:
- Java version of https://github.com/xpring-eng/Xpring-JS/pull/330
- **Overrides #174**
- Add `IlpException` for all Ilp related exceptions
- Translate grpc `StatusRuntimeException`s to `IlpException`
- Clean up tests

### Context of Change
Errors that we get back from gRPC services aren't all that informative or easy to work with.  We also want to hide the network layer from end users, and rethrowing gRPC errors doesn't accomplish this.  By translating gRPC errors into native `IlpException`s, we have more control over what the end user is dealing with and can provide a richer error model for them.

### Type of Change

## Before / After
`IlpClient.getBalance` and `IlpClient.sendPayment` would throw `StatusRuntimeException`s if something went wrong
Now they throw `IlpException`s

## Test Plan
Added a ton of tests


src/main/java/io/xpring/ilp/DefaultIlpClient.java
```diff
@@ -6,16 +6,17 @@
 import org.interledger.spsp.server.grpc.IlpOverHttpServiceGrpc;
 import org.interledger.spsp.server.grpc.SendPaymentRequest;
 import org.interledger.spsp.server.grpc.SendPaymentResponse;
+
```
keefertaylor, line 9: fancy! I promise I will set up automated linting soon.

src/main/java/io/xpring/ilp/IlpException.java
```diff
@@ -0,0 +1,42 @@
+package io.xpring.ilp;
+
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
+
+public class IlpException extends Exception {
+
+  public static final IlpException NOT_FOUND = new IlpException(IlpExceptionType.ACCOUNT_NOT_FOUND, "Account not found.");
```
keefertaylor, line 8: super nit: alphabetize these names?

src/main/java/io/xpring/ilp/IlpException.java
```diff
@@ -0,0 +1,42 @@
...
+  );
+
+  private IlpExceptionType type;
+
+  public IlpException(IlpExceptionType type, String message) {
+    super(message);
+    this.type = type;
+  }
+
+  public IlpExceptionType getType() {
```
keefertaylor, line 25: I'm almost positive that CI will yell at you that you need to doc these methods? At the very least could you javadoc the constructor and the from method?

src/main/java/io/xpring/ilp/IlpExceptionType.java
```diff
@@ -0,0 +1,10 @@
+package io.xpring.ilp;
+
+public enum IlpExceptionType {
+  INVALID_ACCESS_TOKEN,
```
keefertaylor, line 4: Alphabetize these cases

PR #178, opened 2020-04-03
Add `network` parameter in `XRPClient`

keefertaylor:

Add a network parameter and associated getter in `XRPClient`, `XRPClientInterface` and `FakeXRPClient`.

### Context of Change

A future PR will use this network parameter in a Xpring Client which composes XRP and PayID.

This PR also refactors `XRPLNetwork` from the Pay ID package to a common package, since it is now used between XRP and PayID.

Note this is a breaking change. We're already going to do a major semver release to drop the legacy protocol buffer compatibility so we're already going to break all clients of this constructor so we can rip the bandage off all at once.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

No change - client's must update constructor.

## Test Plan

CI


README.md
```diff
@@ -155,22 +154,24 @@ wallet.verify(message, signature); // true
 `XRPClient` is a gateway into the XRP Ledger. `XRPClient` is initialized with a single parameter, which is the URL of the remote adapter (see [Server-Side Component][#server-side-component]).
```
amiecorso, line 154: should this be updated to reflect the fact that an XRPClient is initialized with more than a single param?

src/main/java/io/xpring/xrpl/XRPClient.java
```diff
@@ -1,5 +1,8 @@
 package io.xpring.xrpl;
 
+import io.xpring.common.XRPLNetwork;
```
amiecorso, line 3: duplicate import?

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - All legacy services are removed from XpringKit. All RPC's go through [rippled's protocol buffer API](https://github.com/ripple/rippled/pull/3254).
 
+#### Changed
+- `XRPClient` requires a new parameter in it's constructor that identifies the network it is attached to.
```
intelliot, line 17:
```suggestion
- `XRPClient` constructor requires a new parameter that identifies the network it is connected to.
```

I was going to suggest `it's` -> `its` but this rewording reads nicer, imho

README.md
intelliot: I echo @amiecorso

PR #179, opened 2020-04-03
XpringClient Implementation and Tests

keefertaylor:
## High Level Overview of Change

Create a new class, `XpringClient`, which composes sub-services provided by Xpring.

### Context of Change

`XpringClient` will compose common interactions across Xpring.

Analog in JS: https://github.com/xpring-eng/Xpring-JS/pull/282

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

## Before / After

Easy `XpringClient` usage.

## Test Plan

New tests provided for CI.

PR #181, opened 2020-04-03
Add a `FakeWallet`

keefertaylor:

Create a `FakeWallet` which always produces the given signature.

### Context of Change

This is needed to properly unit test some of the upcoming Xpring functionality.

Note that I only need to partially fake some of the functionality in this class, so I opted to extend `Wallet` and subclass, rather than writing a `Wallet` interface and perfectly hermetic fake. This is kind of lazy, but also pragmatic and we can always refactor later if needed. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

N/A
## Test Plan

CI


keefertaylor:
Argh, actually this isn't quite ready for review yet. Sorry about that.

keefertaylor:
Correct. #179 if you'd like to follow along :)

PR #182, opened 2020-04-03
Add initializer for `Wallet` from public and private keys

keefertaylor:

Add an initializer that lets users create a wallet from public and private keys.

### Context of Change

Create an initializer and unit tests that work on raw public / private keys. Note that the signature:
```
public Wallet(String, String, boolean)
```
is already consumed by:
```
public Wallet(String mnemonic, String derivationPath, boolean isTest)`
```

So this method gets added as a static member. I'm not sure if there's a better pattern. 

Wire tests for this method as well. Note that the underlying JS implementation doesn't do any validity checking on the input keys so there's actually no way to make this method throw. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Can construct wallets from public / private keys. This is a useful feature to have, but also lets us create a `FakeWallet` class for testing.

## Test Plan

New unit tests introduced for CI.


keefertaylor:
I'm mostly interested in @sappenin  has any architectural guidance here on deduping these methods. 

On one hand, the correct answer is probably to make a WalletFactory, but that's a large refactor. This change doesn't preclude that from happening either, so I think it's fine, but I'd value the input of someone who has spent more time thinking about patterns in java.

PR #193, opened 2020-04-07
Enable Linter and Fix Lint Errors

keefertaylor:

We had a CheckStyle config that didn't work locally and didn't complain or verify itself. Enable that. 

Fix resulting lint errors. 

### Context of Change

I'm sorry for the long PR. There's no good way to do this in pieces. Changes introduced in code are recommended by the linter and are not changing functionality. Reviewers likely want to look at:
- `pom.xml`
- `CONTRIBUTING.md`

The configuration is:
- Run lint checks on `verify` build cycle phase (thus developers can happily run `mvn compile` or `mvn test` locally without lint errors)
- Run lint in a strict configuration that will error on warnings and errors
- CI runs `mvn install` which includes the `verify` build phase and thus runs lint

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Before: Ugly code
After: Pretty code, automatically enforced by a small computer program 

## Test Plan

CI proves that lint errors are resolved and that tests continue to pass.

### DIFF
--- a/src/main/java/io/xpring/common/XRPLNetwork.java
+++ b/src/main/java/io/xpring/common/XRPLNetwork.java
@@ -3,18 +3,19 @@ package io.xpring.common;
 /**
  * Possible networks to resolve.
  */
+@SuppressWarnings("checkstyle:AbbreviationAsWordInName")
 public enum XRPLNetwork {
-    DEV("devnet"),
-    TEST("testnet"),
-    MAIN("mainnet");
+  DEV("devnet"),
+  TEST("testnet"),
+  MAIN("mainnet");
 
-    private String networkName;
+  private String networkName;
 
-    XRPLNetwork(String networkName) {
-        this.networkName = networkName;
-    }
+  XRPLNetwork(String networkName) {
+    this.networkName = networkName;
+  }
 
-    public String getNetworkName() {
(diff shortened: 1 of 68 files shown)


src/main/java/io/xpring/common/XRPLNetwork.java
```diff
@@ -3,18 +3,19 @@
 /**
  * Possible networks to resolve.
  */
+@SuppressWarnings("checkstyle:AbbreviationAsWordInName")
```
hbergren, line 6: What warning is this? I see this in a lot of files, can you just disable whatever rule this is?

keefertaylor: Oh sorry, my PR description should have mentioned this. I filed https://app.asana.com/0/1143722888069470/1170259834689737 to track. 

This is yelling at me to not abbreviate things in names. It wants the classes to be named `XrpClient` instead of `XRPClient` or `PayId` rather than `PayID`. 

I still haven't decided on a standard yet, and changing class names (and or changing iVar names inside of classes inconsistently) feels bad so I'm basically just deferring the problem, for now. 

I'd like this to not get worse though, so I've left the rule enabled globally.

PR #196, opened 2020-04-10
New PayID Format Support 

keefertaylor:

Migrate PayID format from:
`$xpring.money/georgewashington`
to
`georgewashington$xpring.money`

### Context of Change

Product requested change in format. 

Remove references to 'payment pointer' and rename a few classes / methods to reflect that we're not longer compliant with the payment pointer spec.

This is a breaking change, but PayID is experimental so I don't consider it a big deal.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

New spec implemented.

## Test Plan

New tests provided for CI. 

## Future Tasks
Uptake parsing logic and test cases in client libraries (JS / Swift / Java)


src/test/java/io/xpring/payid/PayIDIntegrationTest.java
```diff
@@ -33,7 +31,7 @@ public void testXRPAddressForPayIDKnownAddressMainnet() throws PayIDException {
   public void testXRPAddressForPayIDKnownAddressTestnet() throws PayIDException {
     // GIVEN a Pay ID that will resolve on Testnet.
     PayIDClient payIDClient = new PayIDClient(XRPLNetwork.TEST);
-    String payID = "$dev.payid.xpring.money/hbergren";
+    String payID = "alice$dev.payid.xpring.money";
```
austin-king, line 34: We ran into some friction with this given that this is a dependency not only upon the reference implementation, but also with deploying the reference implementation to the specific VM behind this URL. Not asking you to change this, but curious to hear what you think the ideal solution would be here. Like spin up a local database that resolves this address? Not sure, feel like you'd have better perspective on this

keefertaylor: Ideally we'd spin up some sort of docker container that has hermetic data. 

I was generally waiting for things to stabilize before undertaking that endeavor. I think the ILP team has some framework they use for this. For XRP, we've been just using testnet accounts which have been relatively stable. 

This is what I feel is the most pragmatic approach, given that we're time constrained.

austin-king: Cool, ya I don't think it's mission critical that we address it now. Just wanted to learn best practices on this. Thanks!

src/main/java/io/xpring/payid/javascript/JavaScriptPayIDUtils.java
```diff
@@ -41,8 +41,8 @@ public JavaScriptPayIDUtils() throws JavaScriptLoaderException {
   public PayIDComponents parsePayID(String payID) {
     Objects.requireNonNull(payID);
 
-    Value parsePayIDFunction = javaScriptPayIDUtils.getMember("parsePaymentPointer");
-    Value javaScriptComponents = parsePayIDFunction.execute(payID);
+    Value parsePayIDFunction = javaScriptPayIDUtils.getMember("parsePayID");
+    Value javaScriptComponents =  parsePayIDFunction.execute(payID);
```
hbergren, line 45: nit: extra space (Does Java have a linter? or Java-Prettier?)

keefertaylor: Fixed. #193 enabled the linter under pretty strict rules so you would think that it would complain. 

I don't feel like I have time at the moment to really dig into this; but there is lint support for the project and I did fix the spacing.

hbergren: No worries, just seemed like the sort of nit that could potentially be automated.

PR #199, opened 2020-04-15
Updates XRPTransaction class with additional data fields

amiecorso:
The `XRPTransaction` class is a native SDK class for wrapping protobuf objects.  This PR updates this class to maintain consistency with Xpring-JS by adding the fields `timestamp`, `hash`, and `deliveredAmount`.  This also requires that an `XRPTransaction` object be constructed from a `GetTransactionResponse` protobuf instead of just a `Transaction` protobuf, since these additional fields are found in the metadata of this response object.

The checkstyle loosening will prevent clashes happening due to different default import optimizations in different Java IDEs.

Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/197

### Context of Change
In the process of migrating the Xpring Wallet to rely on the SDK for `paymentHistory` (in `transactionHistoryForWallet`) it was realized that the SDK's notion of a transaction (`XRPTransaction`) was missing some fields that the Wallet's notion of a transaction requires.  These changes were made in `Xpring-JS` and need to be propagated to the other two SDK flavors.

### Type of Change

## Before / After
`XRPTransaction` contains additional fields `hash`, `timestamp`, and `deliveredAmount`

## Test Plan
CI.  Tests have been updated with appropriate fake objects and test values.

## Future Tasks

Do the same for Swift SDK


amiecorso:
Reviewers - 
Keefer the main thing I need help with is the `toHex` function you see in `JavaScriptUtils`.  Originally, I tried to wrap the underlying `toHex` function from `xpring-common-js` in the same fashion that other functions are loaded from that library, but I was having trouble passing the right type to the function... so I temporarily stole a conversion function from the internet and plugged it in there.  I need help understanding the relationship between this Java library and `xpring-common-js`. 

Noah I put you on here because you know Java and SDK.
Ted you've been helping me with the wallet migration (and this reflects those Xpring-JS changes) and continued insight into SDK land doesn't hurt.

src/main/java/io/xpring/xrpl/javascript/JavaScriptUtils.java
```diff
@@ -124,4 +124,21 @@ public String toTransactionHash(String transactionBlobHex) {
     Value hash = transactionBlobToTransactionHashFunction.execute(transactionBlobHex);
     return hash.isNull() ? null : hash.toString();
   }
+
+  /**
+   * Convert the given byte array to a hexadecimal string.
+   *
+   * @param bytes An array of bytes
+   * @return An encoded hexadecimal string.
```
nkramer44, line 132: I like to include what the @param and @return represent so there is no confusion.

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -23,6 +27,11 @@
     return ImmutableXRPTransaction.builder();
   }
 
+  /**
+   * The identifying hash of the transaction.
```
nkramer44, line 31: Need a `@return`

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -23,6 +27,11 @@
...
+   */
+  String hash();
+
   /**
    * The unique address of the account that initiated the transaction.
    */
```
nkramer44, line 37: same here

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -115,15 +124,47 @@
   XRPPayment paymentFields();
 
   /**
-   * Constructs an {@link XRPTransaction} from a {@link Transaction}.
+   * The timestamp of the transaction reported in Unix time (seconds).
+   *
+   * @see "https://xrpl.org/basic-data-types.html#specifying-time"
```
nkramer44, line 129: same

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -115,15 +124,47 @@
...
+  Integer timestamp();
+
+  /**
+   * (Omitted for non-Payment transactions) The Currency Amount actually received by the Destination account.
+   * Use this field to determine how much was delivered, regardless of whether the transaction is a partial payment.
+   *
+   * @see "https://xrpl.org/transaction-metadata.html#delivered_amount"
+   */
+  @Nullable
+  String deliveredAmount();
```
nkramer44, line 141: I would recommend typing this as `Optional<String>` instead of making it `@Nullable`

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -115,15 +124,47 @@
...
+   *
+   * @see "https://xrpl.org/basic-data-types.html#specifying-time"
+   */
+  @Nullable
+  Integer timestamp();
+
+  /**
+   * (Omitted for non-Payment transactions) The Currency Amount actually received by the Destination account.
+   * Use this field to determine how much was delivered, regardless of whether the transaction is a partial payment.
+   *
```
nkramer44, line 137: need `@return`

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -178,7 +219,38 @@ static XRPTransaction from(Transaction transaction) {
...
+    // the unix epoch. Convert transaction's timestamp to a unix timestamp.
+    // See: https://xrpl.org/basic-data-types.html#specifying-time
+    Integer timestamp;
+    if (getTransactionResponse.hasDate()) {
+      Integer rippleTransactionDate = getTransactionResponse.getDate().getValue();
+      timestamp = rippleTransactionDate != null
+              ? rippleTransactionDate + 946684800
+              : null;
+    } else {
+      timestamp = null;
```
nkramer44, line 232: pretty sure you dont need this `null` assignment.  Uninitialized variables should be `null`.

amiecorso:
Note to self: REBASE TO MASTER BEFORE MERGE

amiecorso:
Thanks for the comments Noah.  I'll take care of that last @return in `toHex` when I verify that function with Keefer.  All of the `Nullables` are gone and replaced with `Optionals`! Including in the rest of these classes.

keefertaylor:
Hex functions just live on the Utils class:
https://github.com/xpring-eng/xpring4j/blob/master/src/main/java/io/xpring/xrpl/Utils.java#L43

I basically didn't want to traverse down to the JavaScript VM to do hex/bin conversions since that was mostly a solved problem and there were a bunch of stack over flow answers readily available.

amiecorso:
> Hex functions just live on the Utils class:
> https://github.com/xpring-eng/xpring4j/blob/master/src/main/java/io/xpring/xrpl/Utils.java#L43
> 
> I basically didn't want to traverse down to the JavaScript VM to do hex/bin conversions since that was mostly a solved problem and there were a bunch of stack over flow answers readily available.

Awesome, thanks Keefer.  Took care of this.

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -201,7 +235,34 @@ static XRPTransaction from(Transaction transaction) {
       }
     }
 
+    // Transactions report their timestamps since the Ripple Epoch, which is 946,684,800 seconds after
+    // the unix epoch. Convert transaction's timestamp to a unix timestamp.
+    // See: https://xrpl.org/basic-data-types.html#specifying-time
+    Optional<Integer> timestamp = Optional.empty();
```
keefertaylor, line 241: I think you want a `long` here since adding 946,684,800 + <arbitrary int value> could overflow. 

The (documentation also notes this](https://xrpl.org/basic-data-types.html#specifying-time):
> Don't convert Ripple Epoch times to UNIX Epoch times in 32-bit variables: this could lead to integer overflows.

keefertaylor: Note that JS has 2^52 values in a `number` so we're not too concerned there: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -201,7 +235,34 @@ static XRPTransaction from(Transaction transaction) {
...
+
+    Optional<String> deliveredAmount = Optional.empty();
+    if (getTransactionResponse.getMeta().hasDeliveredAmount()) {
+      CurrencyAmount currencyAmountDelivered = getTransactionResponse.getMeta().getDeliveredAmount().getValue();
+      switch (currencyAmountDelivered.getAmountCase()) {
+        case XRP_AMOUNT:
+          deliveredAmount = Optional.of(Long.toString(currencyAmountDelivered.getXrpAmount().getDrops()));
+          break;
+        case ISSUED_CURRENCY_AMOUNT:
+          deliveredAmount = Optional.of(Long.toString(currencyAmountDelivered.getXrpAmount().getDrops()));
```
keefertaylor, line 257: this line is the same as 254, assume you meant to get issuedCurrencyAmount? 

We may consider adding a test for this case?

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
amiecorso: wow, great catch.  I updated this to be a `Long`, but just to clarify, should it be a `Number`? or does `Long` work? (I know they both work, but stylistically...)

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
amiecorso: another very good catch.  I added two test cases to check this specifically...

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
keefertaylor: I would use a long since most people will know the precision off the top of their heads. 

The number type I linked above is a JS class. I actually didn't know there was a [Number](https://docs.oracle.com/javase/8/docs/api/java/lang/Number.html) class in java and I certainly have never seen it used.

PR #200, opened 2020-04-16
Add PayID classes for new format

nkramer44:
Add `PayID` and `AbstractPayID` classes for the new PayID format.

### Context of Change
The PayID format was recently changed from Payment Pointer to a URI scheme resembling something like an email address (user$host.tld).  This change is specified in https://github.com/xpring-eng/rfcs/blob/master/payid/src/spec/payid-uri.md.

This PR adds a type safe `PayID` class, as well as an `AbstractPayID` class, which will be generated into immutable classes via the `@Value.Immutable` annotation.  They act as data classes with custom validation for PayID URIs, and will be used by `PayIDClient` to validate PayIDs and will ultimately be used by some PayID resolver to resolve a `PayID` to an HTTP URL.

These classes are rip offs from @sappenin's Java test harness (https://github.com/xpring-eng/rfcs/tree/master/payid/java)

### Type of Change

## Before / After
PayIDs were typed only as `String` in `PayIDClient` and parsed to a `PaymentPointer`. This will change to a `PayID` typed parameter in a future PR, but `PayID` provides a typesafe way of constructing PayIDs.

## Test Plan
There are two test classes provided which test `PayID` construction and validation for an array of valid and invalid PayID `String`s.

- Implement a PayID resolver that implements an upcoming RFC for PayID resolution.
- Update `PayIDClient` to use the new classes and format.

### DIFF
--- a/src/test/java/io/xpring/payid/PayIDValidValuesTest.java
+++ b/src/test/java/io/xpring/payid/PayIDValidValuesTest.java
@@ -143,10 +143,3 @@ public class PayIDValidValuesTest {
     assertThat(upperCasePercentEncoded("foo$%af%2ebar")).isEqualTo("foo$%AF%2Ebar");
   }
 }
-
-// TODO: Ensure no $ in sub-delims?
-// TODO: Ensure that payid always starts with (unreserved / sub-delims)
-// TODO: pct-encoded
-// TODO: Invalid (gen-delims like hash)
-// TODO: normalization (caps, etc).
-// TODO: I18N?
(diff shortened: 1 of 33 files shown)


nkramer44:
+ @tedkalaw because he has the most SDK PayID context besides Keefer
+ @sappenin because he wrote these classes originally

src/test/java/io/xpring/payid/PayIDValidValuesTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+  @Test
+  public void testupperCasePercentEncoded() {
+    assertThat(upperCasePercentEncoded("foo")).isEqualTo("foo");
+    assertThat(upperCasePercentEncoded("Foo$%af%2eBAR")).isEqualTo("Foo$%AF%2EBAR");
+    assertThat(upperCasePercentEncoded("foo$%af%2ebar")).isEqualTo("foo$%AF%2Ebar");
+  }
+}
+
+// TODO: Ensure no $ in sub-delims?
```
sappenin, line 147: I think we can remove these TODO comments. They should all be addressed.

PR #201, opened 2020-04-16
Fix javadocs for linter

nkramer44:

Fix all the javadocs that were breaking the linter.

### Context of Change
Circle is failing because of recent changes to the javadoc linter.  This fixes all the things the linter was complaining about so other PRs can pass circle.

### Type of Change

- [x] Documentation Updates

## Before / After
Bad javadocs.
Good javadocs.

## Test Plan
N/A

### DIFF
--- a/src/main/java/io/xpring/ilp/model/AccountBalance.java
+++ b/src/main/java/io/xpring/ilp/model/AccountBalance.java
@@ -18,11 +18,16 @@ public interface AccountBalance {
 
   /**
    * The accountId for this account balance.
+   *
+   * @return A {@link String} containing the accountId for this account balance.
    */
   String accountId();
 
   /**
-   * Currency code or other asset identifier that this account's balances will be denominated in.
+   * he currency code or other asset identifier that this account's balances will be denominated in.
+   *
+   * @return A {@link String} containing the currency code or other asset identifier that this account's
+   *          balances will be denominated in.
    */
   String assetCode();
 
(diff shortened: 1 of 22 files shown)


nkramer44:
@amiecorso I left a couple TODOs in here.  A bunch of the immutable interfaces have `@Nullable` fields.  This is fine, but I would opt for using `Optional` instead. Makes code much more readable but in declaration and at the call site. 

Sorry I didn't catch that in your previous PRs!

src/main/java/io/xpring/ilp/model/AccountBalance.java
```diff
@@ -18,11 +18,16 @@
...
   /**
    * The accountId for this account balance.
+   *
+   * @return A {@link String} containing the accountId for this account balance.
    */
   String accountId();
 
   /**
-   * Currency code or other asset identifier that this account's balances will be denominated in.
+   * he currency code or other asset identifier that this account's balances will be denominated in.
```
theotherian, line 27: Oh lawd he comin

nkramer44: image

src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -34,17 +36,24 @@
    * If provided, this transaction is only valid if the sending account's previously-sent transaction matches the
    * provided hash.
    * </p>
+   *
+   * @return A byte array containing the has value of another transaction.
```
theotherian, line 40: Can haz hash value?

PR #203, opened 2020-04-18
Updates the XRP component proto wrappers to use Optionals instead of Nullables

amiecorso:
  This PR replaces optional fields in the wrapper classes that were decorated as `@Nullable` to simply be `Optional` types, and updates the corresponding syntax around handling and testing these values. Use of `null` in Java is dangerous because it poses the risk of easily overlooked `NullPointerException`s and leads to verbose and error-prone non-null checks.  Introduced in Java 8, the [`Optional` type](https://www.oracle.com/technical-resources/articles/java/java8-optional.html) is now the idiomatic way to represent values that may or may not be present, leading to cleaner and more self-explanatory code.

Note that sequence types that may be optional do not use `Optional`, but are allowed to be empty.  Per Neil's advice [here](https://ripplelabs.slack.com/archives/GNJSQ70KB/p1587321027003200).

Documentation: I have no idea how the html tags got into the javadocs in the first place, but this removes them.

Additional small style cleanups implemented now that I have a getter grasp of the Java protobufs and how they're intended to be used. 

### Context of Change
Maintain the SDKs at a high standard and keep them as idiomatic as possible.

### Type of Change
This is a potentially breaking change for applications that directly use SDK objects, as values that were previously bare and nullable must now be unwrapped as `Optional`s.

## Before / After
No potentially `null` fields in the protobuf wrapper classes. 

## Test Plan
CI.  Tests have been updated to unwrap `Optional`s where necessary.


src/main/java/io/xpring/xrpl/model/XRPMemo.java
```diff
@@ -19,33 +17,26 @@
...
-   * Hex value representing characters allowed in URLs.
-   * <p>
+   * (Optional) Hex value representing characters allowed in URLs.
    * Conventionally containing information on how the memo is encoded, for example as a MIME type.
-   * </p>
    *
    * @return A byte array of arbitrary hex value containing the characters allowed in URLs.
    */
-  @Nullable
   byte[] format();
```
nkramer44, line 32: I'm honestly not sure what will happen here if you build an `XRPMemo` without specifying a `format`.  It might blow up and say that you have to set it.  I would test this out to confirm, and if true, do something like so:
```java
@Value.Default
default byte[] format() {
    return new byte[0];
}
```

src/main/java/io/xpring/xrpl/model/XRPMemo.java
```diff
@@ -19,33 +17,26 @@
...
-   * Hex value representing characters allowed in URLs.
-   * <p>
+   * (Optional) Hex value representing characters allowed in URLs.
    * Conventionally, a unique relation (according to RFC 5988) that defines the format of this memo.
-   * </p>
    *
    * @return A byte array of arbitrary hex value containing the characters allowed in URLs.
    */
-  @Nullable
   byte[] type();
```
nkramer44, line 40: same here

src/main/java/io/xpring/xrpl/model/XRPPayment.java
```diff
@@ -38,35 +38,29 @@
    *
    * @return An {@link Integer} containing the tag that identifies the reason for the payment.
```
nkramer44, line 39: update this for Optional

src/main/java/io/xpring/xrpl/model/XRPPayment.java
```diff
@@ -38,35 +38,29 @@
...
    */
-  @Nullable
-  Integer destinationTag();
+  Optional<Integer> destinationTag();
 
   /**
    * (Optional) Minimum amount of destination currency this transaction should deliver.
    *
    * @return An {@link XRPCurrencyAmount} representing the minimum amount of destination currency this
```
nkramer44, line 46: same here + everywhere

src/main/java/io/xpring/xrpl/model/XRPPayment.java
```diff
@@ -38,35 +38,29 @@
...
-  XRPCurrencyAmount deliverMin();
+  Optional<XRPCurrencyAmount> deliverMin();
 
   /**
    * (Optional) Arbitrary 256-bit hash representing a specific reason or identifier for this payment.
    *
    * @return A byte array containing a 256-bit hash representing a specific reason or identifier for this payment.
    */
-  @Nullable
   byte[] invoiceID();
```
nkramer44, line 56: default here

src/main/java/io/xpring/xrpl/model/XRPPayment.java
```diff
@@ -38,35 +38,29 @@
...
-   * Array of payment paths to be used for this transaction.
-   * <p>
+   * (Optional) Array of payment paths to be used for this transaction.
    * Must be omitted for XRP-to-XRP transactions.
-   * </p>
    *
    * @return A {@link List} of {@link XRPPath}s containing the paths to be used for this transaction.
    */
-  @Nullable
   List<XRPPath> paths();
```
nkramer44, line 64: also probably want to default to an empty `List` here

src/main/java/io/xpring/xrpl/model/XRPPathElement.java
```diff
@@ -58,10 +55,25 @@
    * PathElement protocol buffer</a>
    */
   static XRPPathElement from(PathElement pathElement) {
+    Optional<String> account = Optional.empty();
+    if (pathElement.hasAccount()) {
+      account = Optional.of(pathElement.getAccount().getAddress());
```
nkramer44, line 60: is it possible that `.getAddress()` will return `null`?  If so, you should change to 
```java
Optional.ofNullable(pathElement.getAccount().getAddress());
```

You have this control logic in a bunch of different places, so I'd suggest making that change everywhere

src/test/java/io/xpring/xrpl/ProtocolBufferConversionTest.java
```diff
@@ -347,17 +345,16 @@ public void convertPaymentTransactionWithOnlyMandatoryCommonFieldsSetTest() {
...
+    assertThat(xrpTransaction.flags().isPresent()).isFalse();
+    assertThat(xrpTransaction.lastLedgerSequence().isPresent()).isFalse();
+    assertThat(xrpTransaction.memos()).isEmpty();
     assertThat(xrpTransaction.sequence()).isEqualTo(transactionProto.getSequence().getValue());
-    assertThat(xrpTransaction.signers()).isNull();
+    assertThat(xrpTransaction.signers()).isEmpty();
     assertThat(xrpTransaction.signingPublicKey())
         .isEqualTo(transactionProto.getSigningPublicKey().getValue().toByteArray());
-    assertThat(xrpTransaction.sourceTag()).isEqualTo(SourceTag.newBuilder().build().getValue());
+    assertThat(xrpTransaction.sourceTag().isPresent()).isFalse();
```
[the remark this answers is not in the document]
nkramer44: same everywhere

src/main/java/io/xpring/xrpl/model/XRPCurrencyAmount.java
```diff
@@ -54,9 +48,8 @@ static XRPCurrencyAmount from(CurrencyAmount currencyAmount) throws NumberFormat
     switch (currencyAmount.getAmountCase()) {
       // Mutually exclusive: either drops or issuedCurrency is set in an XRPCurrencyAmount
       case ISSUED_CURRENCY_AMOUNT: {
-        IssuedCurrencyAmount issuedCurrencyAmount = currencyAmount.getIssuedCurrencyAmount();
-        if (issuedCurrencyAmount != null) {
-          XRPIssuedCurrency xrpIssuedCurrency = XRPIssuedCurrency.from(issuedCurrencyAmount);
+        if (currencyAmount.hasIssuedCurrencyAmount()) {
```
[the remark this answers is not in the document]
amiecorso: From what I've noticed digging around in the protobuf code for Java, they're designed to not ever return null... instead they return default empty instances of things and handle null things internally.  I'm still not entirely sure what this means when it comes to the value of primitives such as `int` or `string` ... I know empty strings are returned in the absence of value. 

Anyway, yes these boolean `hasWhateverField` methods are a way of checking whether the field in question has been set.  In this case, I actually just deleted this line because it's been replaced with a switch/case statement and I forgot to delete it.

src/main/java/io/xpring/xrpl/model/XRPMemo.java
amiecorso: It will blow up.  However, the code that constructs an `XRPMemo` is in its `.from` method, which always remembers to set these fields... and even if they're not set on the protobuf object, we get a value such as `ByteString.EMPTY` which ends up constructing an empty `byte[]`.  However, I added the defaults for readability since there are a few layers to that logic that are not obvious at a glance.

src/main/java/io/xpring/xrpl/model/XRPPayment.java
amiecorso: added default behavior to any optional sequence type

src/test/java/io/xpring/xrpl/ProtocolBufferConversionTest.java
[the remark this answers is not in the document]
amiecorso: replaced everywhere

src/main/java/io/xpring/xrpl/model/XRPPayment.java
amiecorso: Updated javadocs with Optionals everywhere.

src/main/java/io/xpring/xrpl/model/XRPPathElement.java
amiecorso: So the reason I didn't use this type of logic is because the Java protobufs don't return `null`... so what would instead happen is we'll get an `Optional` object that has some sort of value such as an empty string, when what we really want is an `Optional` that is truly empty and so can be handled as such.

checkstyle.xml
```diff
@@ -164,12 +164,6 @@
         </module>
         <module name="OverloadMethodsDeclarationOrder"/>
         <module name="VariableDeclarationUsageDistance"/>
-        <module name="CustomImportOrder">
```
keefertaylor, line 167: Not opposed, but why remove this?

src/main/java/io/xpring/xrpl/model/XRPMemo.java
```diff
@@ -19,34 +17,36 @@
   }
 
   /**
-   * An arbitrary hex value, conventionally containing the content of the memo.
+   * (Optional) An arbitrary hex value, conventionally containing the content of the memo.
```
keefertaylor, line 20: Is this 'optional'? It just returns a 0 byte if there's no data?

checkstyle.xml
amiecorso: I removed this because I just couldn't otherwise figure out how to get the checkstyle plugin to calm down.  I tried changing the settings on IntelliJ's import ordering so that the program would automatically perform the reorderings across all the files to no avail, and also tried manual reorderings and couldn't satisfy the checkstyler.  At a certain point I had wasted way too much time on it.  Maybe we could solve this in a separate PR dedicated specifically to getting the import settings worked out how we want them?  I'm sure there's a happy middleground that gets us some alphabetizing without being overly strict.

src/main/java/io/xpring/xrpl/model/XRPMemo.java
amiecorso: So this is optional in the sense that it doesn't have to be present, but it's not an `Optional` type because according to Neil you don't typically use `Optional`s in conjunction with `Collection`s, you just return an empty `Collection`... Noah suggested the default byte array values - and `new byte[0]` returns an empty byte array of length 0, not the 0 byte.

keefertaylor: Ah, I read this as an array with a single value, 0x00. Sure, this makes sense.

checkstyle.xml
keefertaylor: Would be great to follow up. 

I believe what this is saying is create groups that are:

`static import`

special groups (org.interledger) - this is the wrong value and is a valuable lesson in stealing configs from other projects

third party packages - not sure

standard java packages. 

Each group is broken out by a vertical space and in each group imports are alphabetized. This isn't a huge deal but if that clears things up and we can avoid removing this rule that would be great.

amiecorso: Ok great, I've assigned myself an Asana task for this next sprint.

src/main/java/io/xpring/xrpl/model/XRPPathElement.java
nkramer44: Ah right, I forgot protobufs have default empty values.

So what will happen here is that if `pathElement.getAccount().getAddress()` returns `""`, then `Optional.of` that will not produce an empty `Optional`.  Maybe I'm missing something but it should probably be 
`Optional.ofNullable(pathElement.getAccount().getAddress().isEmpty() ? null : pathElement.getAccount().getAddress())`

PR #204, opened 2020-04-20
Add WebFinger Response Objects

nkramer44:

Add two immutable class definitions representing a JSON Resource Descriptor (JRD) for WebFinger requests.

### Context of Change
PayID dereferencing will now be accomplished using the WebFinger standard as defined [here](https://tools.ietf.org/html/rfc7033).  More details about this flow can be found in the [PayID Discovery RFC](https://github.com/xpring-eng/rfcs/blob/master/payid/src/spec/payid-discovery.md).

We will need these classes to deserialize JRDs to dereference PayIDs.

### Type of Change

## Before / After
New JRD classes.

## Test Plan
Two JSON serialization/deserialization tests.

Use these to actually dereference a PayID

### DIFF
--- a/src/main/java/io/xpring/payid/WebFingerJrd.java
+++ b/src/main/java/io/xpring/payid/WebFingerJrd.java
@@ -9,8 +9,10 @@ import java.util.List;
 /**
  * Represents a JSON Resource Descriptor (JRD) returned from a WebFinger request as defined in RFC-7033.
  *
+ * <p>
  * For the purposes of PayID dereferencing, only subject and links are necessary.  Thus, the aliases and properties
  * fields are omitted.
+ * </p>
  *
  * @see "https://tools.ietf.org/html/rfc7033#section-4.4"
  */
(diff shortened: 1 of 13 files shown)


src/main/java/io/xpring/payid/WebFingerLink.java
```diff
@@ -0,0 +1,47 @@
...
+   * @return A {@link String} representing the target URI.
+   */
+  String href();
+
+  /**
+   * The media type of what the result of dereferencing the link should be.
+   *
+   * @return A {@link String} representing the media type.
+   */
+  String type();
```
sappenin, line 45: This can be removed.

src/main/java/io/xpring/payid/WebFingerLink.java
```diff
@@ -0,0 +1,47 @@
...
+   * @return A {@link String} representing the relation type.
+   */
+  String rel();
+
+  /**
+   * The target URI of a {@link WebFingerLink}.
+   *
+   * @return A {@link String} representing the target URI.
+   */
+  String href();
```
sappenin, line 38:
```suggestion
  Optional<String> href();
  Optional<String> template();
```

src/main/java/io/xpring/payid/WebFingerJrd.java
```diff
@@ -0,0 +1,55 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.immutables.value.Value;
+
+import java.util.List;
+
+/**
+ * Represents a JSON Resource Descriptor (JRD) returned from a WebFinger request as defined in RFC-7033.
+ *
+ * <p>
+ * For the purposes of PayID dereferencing, only subject and links are necessary.  Thus, the aliases and properties
```
sappenin, line 13:
```suggestion
 * For the purposes of PayID Discovery, only subject and links are necessary.  Thus, the aliases and properties
```

nkramer44:
Closing for now, since this seems to be out of scope for May 7th.

PR #206, opened 2020-04-20
Implement Automode PayID Dereferencing

nkramer44:
Implement a `PayIDResolver` which dereferences a PayID via WebFinger, as described [here](https://github.com/xpring-eng/rfcs/blob/master/payid/src/spec/payid-discovery.md#automated-mode)

### Context of Change
PayIDs will be resolved to URLs by trying to do automode, but will fall back to manual mode if that fails.

WebFinger Links can include an href with another WebFinger URL, so this implementation recursively performs GETs on WebFinger URLs until a PayID server URL is found.

This implementation isn't used anywhere yet.

### Type of Change

## Before / After
New automode PayID resolution implementation.

## Test Plan
Wrote a bunch of tests in `AutoModePayIDResolverTest`.  The `AutoModePayIDResolver` class encapsulates the HTTP call into its own method so that we can easily mock the network call using Mockito.

## Future Tasks
Find a URL Template library that is a more lightweight dependency than spring-web


nkramer44:
Closing for now, since this seems to be out of scope for May 7th.

PR #214, opened 2020-04-29
Abstract XRP Specifics of PayIDClient to subclass

keefertaylor:

`PayIDClient` should be able to support more networks that just XRP. Abstract the specifics to a subclass. 

### Context of Change

Note this is mostly a subclass and rename PR. The only substantiative change in this PR is changing `PayIDClient`'s constructor from taking an `XRPLNetwork` enum to an arbitrary `String`. 

A future set of PRs will:
- Wire up tests proving PayIDClient works with a BTC address
- Generalize the return types in PayIDClient
- Ensure that `XRPPayIDClient` always returns an X-Address

Note this change is breaking but PayID is experimental.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

N/A - No feature changes yet.

## Test Plan

Tests updated for CI.


src/main/java/io/xpring/payid/XRPPayIDClient.java
```diff
@@ -0,0 +1,34 @@
...
+   * Construct a new XRPPayIDClient.
+   *
+   * @param xrplNetwork The XRP Ledger network that this client attaches to.
+   */
+  public XRPPayIDClient(XRPLNetwork xrplNetwork) {
+    super("xrpl-" + xrplNetwork.getNetworkName());
+
+    this.xrplNetwork = xrplNetwork;
+  }
+}
```
amiecorso, line 34: should this class implement `xrpAddressForPayID` ?

amiecorso: (and PayIDClient should have a method just called `addressForPayID` ? )  Or maybe all happening in subsequent PRs?

keefertaylor: It will in the future.

PR #217, opened 2020-04-30
Adds `xrpToDrops` and `dropsToXrp` to XRP/Utils

amiecorso:
This PR adds the utility functions `xrpToDrops` and `dropsToXrp` to `XRP/xrp-utils.ts`.  These utility methods originally existed in `ripple-lib`, which is a robust and well-tested implementation that can handle either string or numeric values.  Their implementation here is a direct translation into Java.  Note, however, that the Java methods can currently only handle String arguments. Test cases are also based on the cases included in `ripple-lib` and added to `xrpl/helpers/UtilsTest`.
 
The logic in the conversions begins with a regex to determine whether the argument is a well-formed number, and includes additional checks such as (in `dropsToXrp`) that any drops amount submitted as a decimal is actually a whole number (i.e. 5.00, 0.0 etc.) and (in `xrpToDrops`) that any xrp amount submitted as a decimal doesn't contain more than 6 numbers after the decimal, as the 7th and beyond would represent a fractional amount in drops - not allowed.

### Context of Change
Since this is such a common conversion, a robust implementation should be part of the SDKs. 

### Type of Change

## Before / After
`xrpToDrops` and `dropsToXrp` can now be imported from `xpring4j` `Utils` class

## Test Plan
New test cases pulled from `ripple-lib` to explore all conversion logic.

## Future Tasks

Implement the same functionality in Swift SDK: https://app.asana.com/0/0/1173726982085828/f

### DIFF
--- a/src/main/java/io/xpring/xrpl/Utils.java
+++ b/src/main/java/io/xpring/xrpl/Utils.java
@@ -126,9 +126,9 @@ public class Utils {
    *
    * @param drops An amount of xrp expressed in units of drops.
    * @return A String representing the drops amount in units of xrp.
-   * @throws Exception if drops is in an invalid format.
+   * @throws IllegalArgumentException if drops is in an invalid format.
    */
-  public static String dropsToXrp(String drops) throws Exception {
+  public static String dropsToXrp(String drops) throws IllegalArgumentException {
     if (drops == null) {
       return null;
     }
@@ -137,10 +137,10 @@ public class Utils {
 
     Matcher dropsMatcher = dropsPattern.matcher(drops);
     if (!dropsMatcher.matches()) {
-      throw new Exception(String.format(
+      throw new IllegalArgumentException(String.format(
               "dropsToXrp: invalid value %s, should be a number matching %s.", drops, dropsRegex));
     } else if (drops.equals(".")) {
-      throw new Exception(String.format(
+      throw new IllegalArgumentException(String.format(
               "dropsToXrp: invalid value %s, should be a BigNumber or string-encoded number.", drops));
     }
 
@@ -156,7 +156,7 @@ public class Utils {
     // This should never happen; the value has already been validated above.
     // This just ensures BigDecimal did not do something unexpected.
     if (!dropsMatcher.matches()) {
-      throw new Exception(String.format(
+      throw new IllegalArgumentException(String.format(
               "dropsToXrp: failed sanity check - value %s does not match %s.", drops, dropsRegex));
     }
 
@@ -168,9 +168,9 @@ public class Utils {
    *
    * @param xrp An amount of xrp expressed in units of xrp.
    * @return A String representing an amount of xrp expressed in units of drops.
-   * @throws Exception if xrp is in invalid format.
+   * @throws IllegalArgumentException if xrp is in invalid format.
    */
-  public static String xrpToDrops(String xrp) throws Exception {
+  public static String xrpToDrops(String xrp) throws IllegalArgumentException {
     if (xrp == null) {
       return null;
     }
@@ -179,10 +179,10 @@ public class Utils {
 
     Matcher xrpMatcher = xrpPattern.matcher(xrp);
     if (!xrpMatcher.matches()) {
-      throw new Exception(String.format(
+      throw new IllegalArgumentException(String.format(
               "xrpToDrops: invalid value, %s should be a number matching %s.", xrp, xrpRegex));
     } else if (xrp.equals(".")) {
-      throw new Exception(String.format(
+      throw new IllegalArgumentException(String.format(
               "xrpToDrops: invalid value, %s should be a BigDecimal or string-encoded number.", xrp));
     }
 
@@ -195,13 +195,13 @@ public class Utils {
     // validated above. This just ensures BigDecimal did not do
     // something unexpected.
     if (!xrpMatcher.matches()) {
-      throw new Exception(String.format(
+      throw new IllegalArgumentException(String.format(
               "xrpToDrops: failed sanity check - value %s does not match %s.", xrp, xrpRegex));
     }
 
     String[] components = xrp.split("[.]");
     if (components.length > 2) {
-      throw new Exception(String.format(
+      throw new IllegalArgumentException(String.format(
               "xrpToDrops: failed sanity check - value %s has too many decimal points.", xrp));
     }
     String fraction = "0";
@@ -209,7 +209,7 @@ public class Utils {
       fraction = components[1];
     }
     if (fraction.length() > 6) {
-      throw new Exception(String.format("xrpToDrops: value %s has too many decimal places.", xrp));
+      throw new IllegalArgumentException(String.format("xrpToDrops: value %s has too many decimal places.", xrp));
     }
     return new BigDecimal(xrp)
             .multiply(new BigDecimal(1000000.0))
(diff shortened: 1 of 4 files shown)


src/main/java/io/xpring/xrpl/Utils.java
```diff
@@ -116,4 +120,97 @@ public static boolean isValidClassicAddress(String address) {
...
+   */
+  public static String dropsToXrp(String drops) throws Exception {
+    String dropsRegex = "^-?[0-9]*['.']?[0-9]*$";
+    Pattern dropsPattern = Pattern.compile(dropsRegex);
+
+    Matcher dropsMatcher = dropsPattern.matcher(drops);
+    if (!dropsMatcher.matches()) {
+      throw new Exception(String.format(
+              "dropsToXrp: invalid value %s, should be a number matching %s.", drops, dropsRegex));
+    } else if (drops == ".") {
```
nhartner, line 139: Better to use `".".equals(drops)` instead of `==`.  See https://stackoverflow.com/questions/767372/string-equals-versus

src/main/java/io/xpring/xrpl/Utils.java
```diff
@@ -116,4 +120,97 @@ public static boolean isValidClassicAddress(String address) {
...
+
+  /**
+   * Convert from units in drops to xrp.
+   *
+   * @param drops An amount of xrp expressed in units of drops.
+   * @return A String representing the drops amount in units of xrp.
+   * @throws Exception if drops is in an invalid format.
+   */
+  public static String dropsToXrp(String drops) throws Exception {
+    String dropsRegex = "^-?[0-9]*['.']?[0-9]*$";
```
nhartner, line 132: First thing should check is that `drops` is not null

src/main/java/io/xpring/xrpl/Utils.java
```diff
@@ -116,4 +120,97 @@ public static boolean isValidClassicAddress(String address) {
...
+    }
+
+    // Converting to BigInteger and then back to string should remove any
+    // decimal point followed by zeros, e.g. '1.00', which is the only valid decimal
+    // representation of drops, which must be whole numbers.
+    // Important: specify base 10 to avoid exponential notation, e.g. '1e-7'
+    drops = new BigDecimal(drops).toBigIntegerExact().toString(10);
+
+    // drops are only whole units
+    if (drops.contains(".")) {
```
nhartner, line 151: I don't see how drops would ever contain a decimal at this point because in the line immediately above `new BigDecimal(drops).toBigIntegerExact()` will never return a floating point number.

src/main/java/io/xpring/xrpl/Utils.java
```diff
@@ -116,4 +120,97 @@ public static boolean isValidClassicAddress(String address) {
...
+   */
+  public static String xrpToDrops(String xrp) throws Exception {
+    String xrpRegex = "^-?[0-9]*['.']?[0-9]*$";
+    Pattern xrpPattern = Pattern.compile(xrpRegex);
+
+    Matcher xrpMatcher = xrpPattern.matcher(xrp);
+    if (!xrpMatcher.matches()) {
+      throw new Exception(String.format(
+              "xrpToDrops: invalid value, %s should be a number matching %s.", xrp, xrpRegex));
+    } else if (xrp == ".") {
```
nhartner, line 181: Change to `xrp.equals(".")`

src/main/java/io/xpring/xrpl/Utils.java
```diff
@@ -116,4 +120,97 @@ public static boolean isValidClassicAddress(String address) {
...
   }
+
+  /**
+   * Convert from units in drops to xrp.
+   *
+   * @param drops An amount of xrp expressed in units of drops.
+   * @return A String representing the drops amount in units of xrp.
+   * @throws Exception if drops is in an invalid format.
+   */
+  public static String dropsToXrp(String drops) throws Exception {
```
nhartner, line 131: Generally it's not common to throw exceptions using the base `Exception` type. You want to pick something more specific. For all the exceptions that you are throwing in these methods,`IllegalArgumentException` would be more appropriate.

src/main/java/io/xpring/xrpl/Utils.java
```diff
@@ -116,4 +120,97 @@ public static boolean isValidClassicAddress(String address) {
...
+  }
+
+  /**
+   * Convert from units in xrp to drops.
+   *
+   * @param xrp An amount of xrp expressed in units of xrp.
+   * @return A String representing an amount of xrp expressed in units of drops.
+   * @throws Exception if xrp is in invalid format.
+   */
+  public static String xrpToDrops(String xrp) throws Exception {
```
nhartner, line 173: Likewise I would change this method to throw `IllegalArgumentException` instead of the more generic `Exception` type.

src/main/java/io/xpring/xrpl/Utils.java
amiecorso: That's correct - this is left over from translating a Javascript implementation which used `BigNumber` instead of `BigInteger` and could have still contained a decimal point.  This logic has been updated by wrapping `toBigIntegerExact()` in a try/catch which will re-throw in case there's a non-whole drops value.

src/main/java/io/xpring/xrpl/Utils.java
```diff
@@ -116,4 +120,96 @@ public static boolean isValidClassicAddress(String address) {
...
+   */
+  public static String dropsToXrp(String drops) throws IllegalArgumentException {
+    Preconditions.checkNotNull(drops);
+
+    String dropsRegex = "^-?[0-9]*['.']?[0-9]*$";
+    Pattern dropsPattern = Pattern.compile(dropsRegex);
+    Matcher dropsMatcher = dropsPattern.matcher(drops);
+
+    if (!dropsMatcher.matches()) {
+      throw new IllegalArgumentException(String.format(
```
keefertaylor, line 139: Non-blocking: 

I had used [`INVALID_INPUTS`](https://github.com/xpring-eng/xpring4j/blob/master/src/main/java/io/xpring/xrpl/XRPExceptionType.java#L5) and `XRPException` to handle cases like this in the past. 

@nhartner what's more idiomatic here? Always to throw the most specific exception? I can revise existing code if so.

src/main/java/io/xpring/xrpl/Utils.java
```diff
@@ -116,4 +120,96 @@ public static boolean isValidClassicAddress(String address) {
...
+    }
+
+    // This should never happen; the value has already been validated above.
+    // This just ensures BigDecimal did not do something unexpected.
+    if (!dropsMatcher.matches()) {
+      throw new IllegalArgumentException(String.format(
+              "dropsToXrp: failed sanity check - value %s does not match %s.", drops, dropsRegex));
+    }
+
+    return new BigDecimal(drops).divide(new BigDecimal(1000000.0)).toPlainString();
```
keefertaylor, line 162: You use this value twice. Would it make it more clear to name it (precision?) or calculate so the reader understands where it's coming from:

scale = 8
precision = Math.pow(10, scale)

src/main/java/io/xpring/xrpl/Utils.java
amiecorso: Maybe I should also be using `XRPException` here because I use the Swift equivalent in the Swift analogue to this PR... probably because Swift doesn't have the same palate of built-in error types like Java does

src/main/java/io/xpring/xrpl/Utils.java
amiecorso: I used a named variable instead of the magic number 1000000.0 to make it more clear.

src/main/java/io/xpring/xrpl/Utils.java
amiecorso: went this route

PR #221, opened 2020-05-04
Fix bug in X-Address Encoding

keefertaylor:

Fix bug where encoding addresses without tags returned incorrect results. 

### Context of Change

The signature of `encodeXAddress` in JavaScript is `address: string, tag: number, isTest: bool`. Without a tag, we were passing `address: string, isTest: bool` which was yielding bad results. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Added a new test case and ensured it failed. Corrected code and ensured it passed. 

## Test Plan

CI

### DIFF
--- a/src/main/java/io/xpring/xrpl/javascript/JavaScriptUtils.java
+++ b/src/main/java/io/xpring/xrpl/javascript/JavaScriptUtils.java
@@ -49,13 +49,18 @@ public class JavaScriptUtils {
    * @return A new X-Address if inputs were valid, otherwise null.
    * @see <a href="https://xrpaddress.info/">https://xrpaddress.info/</a>
    */
+  @SuppressWarnings("checkstyle:LocalVariableName")
   public String encodeXAddress(ClassicAddress classicAddress) {
     Objects.requireNonNull(classicAddress);
 
     Value encodeXAddressFunction = javaScriptUtils.getMember("encodeXAddress");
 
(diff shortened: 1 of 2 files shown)


keefertaylor:
Fishing for a second approval since this blocks some Pay ID stuff.

src/main/java/io/xpring/xrpl/javascript/JavaScriptUtils.java
```diff
@@ -49,14 +49,24 @@ public boolean isValidAddress(String address) {
...
+    if (classicAddress.tag().isPresent()) {
+      Value xAddress = encodeXAddressFunction.execute(
+          classicAddress.address(),
+          classicAddress.tag().get(),
+          classicAddress.isTest()
+      );
+      return xAddress.asString();
+    } else {
+      Value undefined = JavaScriptLoader.getContext().eval("js", "undefined");
+      Value xAddress = encodeXAddressFunction.execute(classicAddress.address(), undefined, classicAddress.isTest());
```
hbergren, line 67: Got it. So the fix here was that you were passing `isTest()` to the `tag` parameter of this function, and now you pass `undefined`?

keefertaylor: Correct. The `isTest` parameter got added in some update to Xpring-JS and I just appended rather than thinking about what was actually happening.

PR #228, opened 2020-05-12
Idiomatically Case `PayIDExceptionType` and `PayIDException`

keefertaylor:

Begin recasing PayID components to be idiomatically cased. 

### Context of Change

Java prefers `Id` to `ID`. This recases the first several classes. 

Because Java only allows one public class per file and MacOS is case insensitive, we can't have these files live side by side. As such, a new package, `idiomatic` is created. We'll migrate things there, free up the original namespace and then migrate them back to avoid breaking clients.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Extra classes added

## Test Plan

CI

### DIFF
--- a/src/main/java/io/xpring/payid/idiomatic/PayIdException.java
+++ b/src/main/java/io/xpring/payid/idiomatic/PayIdException.java
@@ -1,0 +1,38 @@
+package io.xpring.payid.idiomatic;
+
+/**
+ * Represents an exception thrown from PayID Components.
+ */
+public class PayIdException extends Exception {
+  /**
+   * Static exception for when a classic address is passed to an X-Address API.
+   */
+  public static PayIdException invalidPaymentPointerException =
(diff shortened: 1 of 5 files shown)


src/main/java/io/xpring/payid/idiomatic/PayIdException.java
```diff
@@ -0,0 +1,38 @@
+package io.xpring.payid.idiomatic;
+
+/**
+ * Represents an exception thrown from PayID Components.
+ */
+public class PayIdException extends Exception {
```
nhartner, line 6: If `PayIdException` extended `PayIDException` then you could replace usages of `PayIDException` in the SDK with `PayIdException` and legacy clients would still work as expected and get a deprecation warning to catch the new exception instead.

nhartner: Never mind, that won't work because they use a different enum type.

PR #230, opened 2020-05-12
Implements `getPayment` in `XRPClient`

amiecorso:
This PR implements `getTransaction` in `XRPClient` and all related classes, which queries the XRP ledger for a transaction by hash, and if found, returns an `XRPTransaction` object representing the transaction.  Unit and integration tests added.

### Context of Change
Mayur mentioned that a partner (presumably Forte) was looking for this functionality in the SDK. 

### Type of Change

## Before / After
`XRPClient` now exposes `getTransaction`

## Test Plan
New unit and integration tests pass on CI. 

## Future Tasks

Implement same functionality in JS and Swift SDKs


src/main/java/io/xpring/xrpl/DefaultXRPClient.java
```diff
@@ -292,6 +292,28 @@ public boolean accountExists(String address) throws XRPException {
...
+   * Retrieve the transaction corresponding to the given transaction hash.
+   * <p>
+   * Note: This method can return transactions that are not included in a fully validated ledger.
+   *       See the `validated` field to make this distinction.
+   * </p>
+   * @param transactionHash The hash of the transaction to retrieve.
+   * @return An XRPTransaction object representing an XRP Ledger transaction.
+   * @throws XRPException If the transaction hash was invalid.
+   */
+  public XRPTransaction getTransaction(String transactionHash) throws XRPException {
```
keefertaylor, line 306: I think this will only convert payment objects right? 

Maybe rename as `getPayment`?

PR #231, opened 2020-05-14
Updates README with new ILP gRPC URL

amiecorso:
Updates the README to contain correct ILP gRPC URL.

### Type of Change

- [x] Documentation Updates


README.md
```diff
@@ -324,7 +324,7 @@ All calls to `IlpClient` must pass an access token, which can be generated in yo
 ```java
 import io.xpring.ilp.IlpClient;
 
-String grpcUrl = "hermes-envoy-test.xpring.io"; // Testnet Hermes URL
+String grpcUrl = "prod.grpcng.wallet.xpring.io:443"; // Testnet ILP Wallet URL
```
sappenin, line 327: Does the client need the `443` port, or can that be inferred?

keefertaylor: I don't think it can be inferred because this isn't http traffic (though I could be wrong?)

nkramer44: Tested this locally.  Including the port actually doesn't work, so we should get rid of the port here as well

amiecorso: removed

PR #232, opened 2020-05-15
Fix semantics of methods for current open ledger and last validated ledger

keefertaylor:

The current implementation of `getLatestValidatedLedgerSequence` returns the latest *open* ledger sequence. 

Fix this method's name and re-implement `getLatestValidatedLedgerSequence`.

Analogue in Swift: https://github.com/xpring-eng/XpringKit/pull/216

### Context of Change

There are a number of assumptions required to get `getLatestValidatedLedgerSequence` working. Namely, an address that is guaranteed to exist needs to be provided. 

I've worked with @carlhua  and @cjcobb23  on the rippled team and we are convinced that this implementation will never fail to return the latest validated ledger sequence when called from our reliable submission logic. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Reliable submission bug is fixed. 

## Test Plan

CI

### DIFF
--- a/src/main/java/io/xpring/xrpl/DefaultXRPClient.java
+++ b/src/main/java/io/xpring/xrpl/DefaultXRPClient.java
@@ -7,7 +7,24 @@ import io.grpc.StatusRuntimeException;
 import io.xpring.xrpl.model.XRPTransaction;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.xrpl.rpc.v1.*;
+import org.xrpl.rpc.v1.AccountAddress;
+import org.xrpl.rpc.v1.AccountRoot;
+import org.xrpl.rpc.v1.CurrencyAmount;
+import org.xrpl.rpc.v1.GetAccountInfoRequest;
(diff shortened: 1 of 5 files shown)


src/main/java/io/xpring/xrpl/DefaultXRPClient.java
```diff
@@ -292,11 +292,44 @@ public boolean accountExists(String address) throws XRPException {
...
+   * </p>
+   * @param address An address that exists at the current time. The address is unchecked and must be a classic address.
+   * @return The index of the latest validated ledger.
+   * @throws XRPException If there was a problem communicating with the XRP Ledger.
+   */
+  @Override
+  public int getLatestValidatedLedgerSequence(String address) throws XRPException {
+    // rippled doesn't support a gRPC call that tells us the latest validated ledger sequence. To get around this,
+    // query the account info for an account which will exist, using a shortcut for the latest validated ledger. The
+    // response will contain the ledger the information was retrieved at.
```
amiecorso, line 318: I would say "... contain the ledger **index** the information ... " (or "sequence" ... )

PR #236, opened 2020-05-15
Adds `validated` and `ledgerIndex` fields to `XRPTransaction`

amiecorso:
This PR adds the `validated` and `ledgerIndex` fields to the `XRPTransaction` object. 
These fields come from a `GetTransactionResponse` protobuf object, which is returned by a call to `getTransaction`. 

### Context of Change
This change is made in preparation for implementing a `getTransaction` method in the SDKs. The original implementation of the `XRPTransaction` class in the SDK only included the fields present in the `Transaction` protocol buffer object (which is wrapped by a `GetTransactionResponse` also containing other metadata about the transaction's outcome.) Certain of these other metadata fields have been added as-needed to the `XRPTransaction` object, and this PR takes care of the remaining fields available in `GetTransactionResponse`.

Our current approach is to combine the distinct `GetTransactionResponse` and `Transaction` protocol buffer objects into one type (`XRPTransaction`) since the metadata about the transaction's outcome is of equal importance to the specific transaction data, and while the type distinction makes sense in the context of authoring rippled, it would add unnecessary complexity in the SDK. 

### Type of Change

## Before / After
Two additional fields available on an `XRPTransaction` object.

## Test Plan
Added new test values to fake protobuf objects for conversion tests.
CI passes.

## Future Tasks

Update Swift SDK with the same changes.

### DIFF
--- a/src/main/java/io/xpring/xrpl/model/XRPTransaction.java
+++ b/src/main/java/io/xpring/xrpl/model/XRPTransaction.java
@@ -163,6 +163,21 @@ public interface XRPTransaction {
   Optional<String> deliveredAmount();
 
   /**
+   * A boolean indicating whether this transaction was found on a validated ledger, and not an open or closed ledger.
+   *
+   * @return A boolean indicating whether or not this transactionw as found on a validated ledger.
+   * @see "https://xrpl.org/ledgers.html#open-closed-and-validated-ledgers"
+   */
+  boolean validated();
+
+  /**
+   * The index of the ledger on which this transaction was found.
+   *
+   * @return An integer indicating the ledger index of the ledger on which this transaction was found.
+   */
+  int ledgerIndex();
+
+  /**
    * Constructs an {@link XRPTransaction} from a {@link GetTransactionResponse}.
    *
    * @param getTransactionResponse a {@link GetTransactionResponse} (protobuf object) whose field values will be used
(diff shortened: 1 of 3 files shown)


src/main/java/io/xpring/xrpl/model/XRPTransaction.java
```diff
@@ -162,6 +162,21 @@
...
+  /**
+   * A boolean indicating whether this transaction was found on a validated ledger, and not an open or closed ledger.
+   *
+   * @return A boolean indicating whether or not this transactionw as found on a validated ledger.
+   * @see "https://xrpl.org/ledgers.html#open-closed-and-validated-ledgers"
+   */
+  boolean validated();
+
+  /**
+   * The index of the ledger on which this transaction was found.
```
keefertaylor, line 174: I don't think this is always in a "search" context. 

WDYT of: 
```
The ledger index of the ledger in which this transaction was included, or is slated to be included. 
```

Is this field set only if validated == true? Or is it possible to get an open/closed ledger where this value could change?

amiecorso: So I think the `ledger_index` field is set no matter what... this is based on reading the rippled JSON API documentation for the `tx` command, and also looking at the rippled handler code here: https://github.com/ripple/rippled/blob/97712107b71a8e2089d2e3fcef9ebf5362951110/src/ripple/rpc/handlers/Tx.cpp#L253.   So it is possible that the `ledger_index` could ultimately change if it's not being reported from a validated ledger.  

As for the comment change though, I'm not sure that it necessarily clarifies that potential confusion.... it might be easy to mistake "slated to include" as "the final ledger index WILL still be this number, even though it's not validated yet" ... so for now I'm just going to leave the comment in this simple form.

If we really wanted to make the SDK completely foolproof, maybe we would force it to only ever return validated data? Otherwise we leave some of these details up to the knowledge of users.

PR #240, opened 2020-05-19
Rename Additional PayID Objects

keefertaylor:
Re-case some PayID model objections to be idiomatically.

Follow on to https://github.com/xpring-eng/xpring4j/pull/228 

### Context of Change

Java prefers `Id` to `ID`. This recases the first several classes. 

Because Java only allows one public class per file and MacOS is case insensitive, we can't have these files live side by side. As such, a new package, `idiomatic` is created. We'll migrate things there, free up the original namespace and then migrate them back to avoid breaking clients.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Extra classes added

## Test Plan

CI

### DIFF
--- a/src/main/java/io/xpring/payid/PayID.java
+++ b/src/main/java/io/xpring/payid/PayID.java
@@ -16,6 +16,7 @@ import java.util.StringTokenizer;
  *
  * @deprecated Please use the idiomatically named `PayId` interface instead.
  */
+@Deprecated
 public interface PayID {
 
   String PAYID_SCHEME = "payid:";
--- a/src/main/java/io/xpring/payid/idiomatic/PayId.java
+++ b/src/main/java/io/xpring/payid/idiomatic/PayId.java
@@ -1,6 +1,6 @@
-package io.xpring.payid;
+package io.xpring.payid.idiomatic;
 
-import static io.xpring.payid.AbstractPayID.upperCasePercentEncoded;
+import static io.xpring.payid.idiomatic.AbstractPayId.upperCasePercentEncoded;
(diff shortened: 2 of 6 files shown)


src/main/java/io/xpring/payid/idiomatic/PayId.java
```diff
@@ -0,0 +1,106 @@
...
+   */
+  static PayId of(String value) {
+    Objects.requireNonNull(value, "Pay ID must not be null");
+    if (value.toLowerCase(Locale.ENGLISH).startsWith(PAY_ID_SCHEME)) {
+      value = value.substring(6);
+    } else {
+      throw new IllegalArgumentException(format("Pay ID `%s` must start with the 'payid:' scheme", value));
+    }
+
+    if (!value.contains("$")) {
```
Stormtv, line 53: Would it be better to use a regex to validate this?

src/main/java/io/xpring/payid/PayID.java
```diff
@@ -13,6 +13,8 @@
  * A standardized identifier for payment accounts.
  *
  * @see "https://github.com/xpring-eng/rfcs/blob/master/TBD.md"
+ *
+ * @deprecated Please use the idiomatically named `PayId` interface instead.
```
amiecorso, line 17: Why doesn't this class have an `@Deprecated` decorator, but others such as `PayIDComponents` and `AbstractPayID` do?

keefertaylor: Purely an oversight.

PR #250, opened 2020-05-30
Adds X-Addresses to `XRPTransaction` and `XRPPayment` model objects

amiecorso:
This PR adds X-address to the `XRPPayment` and `XRPTransaction` model objects.  `XRPPayment` contains a `destination` and a `destinationTag` property, and now also contains a `destinationXAddress` property that is the X-address encoded version of this information. `XRPTransaction` contains an `account` property, and now also contains an `accountXAddress` property, which is the X-address encoded version of the originating account with no destination tag.

In order to create the correct X-address encoding, we need to know whether we're on Testnet or Mainnet.  Therefore the `.from` constructors for `XRPPayment` and `XRPTransaction` now have an additional `XRPLNetwork` parameter.  Tests and client code have been updated accordingly. 

### Type of Change

## Before / After
`XRPTransaction` and `XRPPayment` now contain X-address versions of address and tag properties. 

## Test Plan
Unit tests for protocol buffer conversion into these objects have been updated to verify the X-address properties.
CI stil passes.

## Future Tasks

Questions for @keefertaylor 
- do we need to deprecate the classic address/tag fields, or just leave everything side-by-side?
- should a note about this go in the CHANGELOG?  The `XRPTransaction` and `XRPPayment` objects have an additional constructor parameter, but it's optional (defaults to Mainnet) and thus isn't a breaking change.
- should other classes that contain XRP addresses (such as `XRPPathElement` or `XRPIssuedCurrency`) also contain X-address versions of these?

### DIFF
--- a/src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
+++ b/src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
@@ -50,8 +50,8 @@ public class FakeXRPProtobufs {
   public static String testCurrencyName = "currencyName";
   public static ByteString testCurrencyCode;
 
-  public static String fakeAddress1 = "r123";
-  public static String fakeAddress2 = "r456";
+  public static String fakeAddress1 = "rsKouRxYLWGseFwXSAo57qXjcGiNqR55wr";
+  public static String fakeAddress2 = "rPuNV4oA6f3SrKA4pLEpdVZW6QLvn3UJxK";
 
   static {
     try {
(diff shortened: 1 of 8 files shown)


src/test/java/io/xpring/xrpl/FakeXRPProtobufs.java
```diff
@@ -50,8 +50,8 @@
   public static String testCurrencyName = "currencyName";
   public static ByteString testCurrencyCode;
 
-  public static String fakeAddress1 = "r123";
-  public static String fakeAddress2 = "r456";
+  public static String fakeAddress1 = "rsKouRxYLWGseFwXSAo57qXjcGiNqR55wr";
+  public static String fakeAddress2 = "rPuNV4oA6f3SrKA4pLEpdVZW6QLvn3UJxK";
```
amiecorso, line 54: I had to update these with valid classic addresses in order to get `Utils.encodeXAddress` to return non-null

PR #256, opened 2020-06-05
Idiomatically Case the Remainder of PayID classes

keefertaylor:

Java prefers to case classes as `PayId` rather than `PayID`.

### Context of Change

This is a one shot PR to re-case all the java classes. Existing classes are deprecated.

Due to Java's requirement that all file names are the same as the class and MacOS's case insensitive file system, we need to place the new classes in a new `idiomatic` package. A future migration will remove this package after the original namespace is freed. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Clients are advised to move to idiomatically named classes.

## Test Plan

CI


CHANGELOG.md
```diff
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `PayIdClient` provides the functionality of `PayIDClient` under an idiomatic cased name.
```
amiecorso, line 11: Perhaps all of these should read "idiomatically" ?

PR #270, opened 2020-06-22
Add Fetch All Addresses Feature and Unbound `PayIDClient` from a specific network

amiecorso:

Adds new functionality for retrieving all addresses. Refactor `PayIDClient` to exclude `network` parameter, new method `cryptoAddressForPayId` replaces `addressForPayId` and requires a network in addition to a PayID. 

Analogue in JS: https://github.com/xpring-eng/Xpring-JS/pull/540
### Context of Change

This change adds a new method to retrieve all addresses for a PayID. Unit / integration tests are also provided. 

There's a larger change here where `PayIDClient` is no longer bound to a network. This is because it doesn't make logical sense, for example, if a `PayIDClient` bound to a BTC network is able to read XRP addresses. As such, this is a breaking change to `PayIDClient`.

Update CHANGELOG and README.md.
Updates around breaking changes in existing tests etc.

### Type of Change

- [x] Documentation Updates

## Before / After
`PayIDClient` can now retrieve all addresses for a PayID and is no longer bound to a specific network.

## Test Plan

CI - new tests are provided (unit + integration)


src/main/java/io/xpring/payid/PayIDClient.java
```diff
@@ -75,13 +58,43 @@ public void setEnableSSLVerification(boolean enableSSLVerification) {
...
+    } else {
+      // With a specific network, exactly one address should be returned by a PayId lookup.
+      throw new PayIDException(PayIDExceptionType.UNEXPECTED_RESPONSE,
+              "Expected one address for " + payId + " on network " + network
+                      + " but got " + addresses.size());
+    }
+  }
+
+  /**
+   * Retrieve the all addresses associated with a PayId.
```
hbergren, line 81:
```suggestion
   * Retrieve the all addresses associated with a PayID.
```

src/main/java/io/xpring/payid/idiomatic/PayIdClient.java
```diff
@@ -19,17 +23,13 @@
 /**
  * Implements interaction with a PayID service.
  * Warning:  This class is experimental and should not be used in production applications.
+ *
```
keefertaylor, line 26: nit: extra line

keefertaylor, line 26:
```suggestion
```

PR #285, opened 2020-07-17
(1/3 java) Refactors and adds new flag enums to support `enableDepositAuth`

amiecorso:
This PR eliminates the `RippledFlags` class and creates a separate enum class for each different type of flag that we use so far in the Xpring SDK.  New flags will support `enableDepositAuth`.  Taking care of this in a separate PR to reduce the change volume in upcoming `enableDepositAuth` PR.

### Context of Change
As the SDKs expand to accommodate more transaction types, we need to thoughtfully expand the associated required flags.

### Type of Change

## Before / After
New flag enum classes:
`PaymentFlag`
`AccountSetFlag`
`AccountRootFlag`

## Test Plan
CI still passes.  New tests added for `AccountRootFlag`.


amiecorso:
@keefertaylor is this considered a breaking change?  Should this change be noted in the Changelog?  or is it sufficiently internal?

PR #299, opened 2020-08-05
Adds Support for Transaction Memos in `XrpClient`

amiecorso:
Adds support for transaction memos via additional public API `sendWithDetails` exposed on `XrpClient` and `XpringClient`.  `sendWithDetails` takes a new object type, a `SendXrpDetails` that bundles the amount, destination, sending wallet, and list of memos.

Analogue in `xpring-js`: https://github.com/xpring-eng/Xpring-JS/pull/501

### Context of Change
The initial implementation of `send` in `XrpClient` was kept intentionally simplistic, however, some business logic (such as ChefXRP) requires the use of the `Memos` option on transactions.  This PR enables that information to be optionally included in a transaction, while also leaving the original simplified API intact.

### Type of Change

- [x] Documentation Updates

## Before / After
Before: memos weren't addable to the transaction
After: memos can be added and created with hex and non-hex strings

## Test Plan
Added unit and integration tests for the new functionality, CI still passes.

## Future Tasks

similar support in Swift


src/test/java/io/xpring/common/CommonUtilsTest.java
```diff
@@ -0,0 +1,60 @@
...
+ * Unit tests for {@link CommonUtils}.
+ */
+public class CommonUtilsTest {
+
+  @Test
+  public void testStringToHexConvertsStringCorrectly() {
+    // GIVEN the best line delivered on film
+    // WHEN that line is converted to hex
+    // THEN the hex value is correct
+    assertEquals(CommonUtils.stringToHex("oh, hi mark"), "6f682c206869206d61726b");
```
tedkalaw, line 18: image

src/main/java/io/xpring/common/CommonUtils.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+    return hex.toString();
+  }
+
+  /**
+   * Converts a string that is optionally in hex format into a byte[].
+   *
+   * @param value - the string to convert to a byte[].
+   * @param isHex - flag to indicate if it's a hex string or not.
+   * @return the byte[] value.
```
tedkalaw, line 25: nit - `The byte[] value`? other `@return`s seem to capitalize the first letter

src/main/java/io/xpring/common/CommonUtils.java
```diff
@@ -0,0 +1,35 @@
+package io.xpring.common;
```
keefertaylor, line 1: These things exist in the poorly named XrpUtils.java: https://github.com/xpring-eng/xpring4j/blob/5c07b0f664493777af211379f748ba2dd5312be0/src/main/java/io/xpring/xrpl/Utils.java.

Agree this should be common. Can we deprecate those methods?

src/main/java/io/xpring/xrpl/model/MemoField.java
```diff
@@ -0,0 +1,18 @@
...
+ * Provides a means of passing a string to a memo that allows for user
+ * specification as to whether or not the string is already a hex string.
+ */
+@Value.Immutable
+public interface MemoField {
+  static ImmutableMemoField.Builder builder() {
+    return ImmutableMemoField.builder();
+  }
+
+  String value();
```
keefertaylor, line 15: Would you mind doc'ing these? I know it's obvious to you since you've been reading the documentation but it may not be to someone without that context.

src/main/java/io/xpring/xrpl/model/SendXrpDetails.java
```diff
@@ -0,0 +1,28 @@
...
+ * destination field may be a PayID, XAddress, or other type of address. Handling
+ * of the given destination type is the responsibility of the client.
+ */
+@Value.Immutable
+public interface SendXrpDetails {
+  static ImmutableSendXrpDetails.Builder builder() {
+    return ImmutableSendXrpDetails.builder();
+  }
+
+  BigInteger amount();
```
keefertaylor, line 21: Same comment re: docs

src/main/java/io/xpring/xrpl/model/MemoField.java
amiecorso: doc'ed

src/main/java/io/xpring/common/CommonUtils.java
amiecorso: deprecated

PR #302, opened 2020-08-10
Creates XrpAccountDelete protobuf wrapper class

tedkalaw:
This PR implements the class XRPAccountDelete - a native Java class to wrap the AccountDelete protocol buffer. See also: https://xrpl.org/accountdelete.html.

### Context of Change

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
XRPAccountDelete class and conversion unit tests now exist.

## Test Plan

### DIFF
--- a/src/test/java/io/xpring/xrpl/XrpAccountDeleteProtoConversionTest.java
+++ b/src/test/java/io/xpring/xrpl/XrpAccountDeleteProtoConversionTest.java
@@ -1,0 +1,55 @@
+package io.xpring.xrpl;
+
+import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
+
+import io.xpring.common.XrplNetwork;
+import io.xpring.xrpl.model.XrpAccountDelete;
+import io.xpring.xrpl.model.XrpAccountSet;
+import org.junit.Test;
+import org.xrpl.rpc.v1.AccountDelete;
+import org.xrpl.rpc.v1.AccountSet;
+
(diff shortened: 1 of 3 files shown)


src/test/java/io/xpring/xrpl/XrpAccountDeleteProtoConversionTest.java
```diff
@@ -0,0 +1,53 @@
...
+  public void destinationTagNotSetTest() {
+    // GIVEN an AccountDelete protocol buffer with no destination tag.
+    final AccountDelete accountDeleteProto = FakeXrpTransactionProtobufs.noDestinationTagAccountDelete;
+
+    // WHEN the protocol buffer is converted to a native Java object.
+    final XrpAccountDelete accountDelete = XrpAccountDelete.from(accountDeleteProto, XrplNetwork.TEST);
+
+    ClassicAddress classicAddress = ImmutableClassicAddress.builder()
+      .address(accountDeleteProto.getDestination().getValue().getAddress())
+      .isTest(true)
```
tedkalaw, line 40: i purposefully didn't use the `destinationTag` function in the builder. should i do this? when i did, it gave a different result, presumably because the default value for `Integer` is 0. not sure how these are used in practice

amiecorso: Yep, I think that's the right thing to do

PR #305, opened 2020-08-11
Implements `randomWalletFromFaucet` for Improved `XrpClient` Integration Tests

amiecorso:
This PR implements the test utility `randomWalletFromFaucet`, which generates a random wallet and then requests funds via the XRPL Testnet faucet (https://xrpl.org/xrp-testnet-faucet.html).  Integration tests are refactored to use a new funded address each time, which should eliminate the failures that occur due to sequence number clashes when different clients (via different CI runs) send transactions on behalf of the same account.

### Type of Change

## Before / After
New addresses are used for integration tests where relevant.

## Test Plan
Refactor integration tests where necessary.  CI still passes.

### DIFF
--- a/src/test/java/io/xpring/xrpl/WalletTest.java
+++ b/src/test/java/io/xpring/xrpl/WalletTest.java
@@ -214,6 +214,6 @@ public class WalletTest {
     // AND it has a non-zero balance
     XrpClient xrpClient = new XrpClient("test.xrp.xpring.io:50051", XrplNetwork.TEST);
     BigInteger balance = xrpClient.getBalance(wallet.getAddress());
-    assertTrue(balance.compareTo(new BigInteger("0")) == 1);
+    assertTrue(balance.signum() == 1);
   }
 }
--- a/src/test/java/io/xpring/xrpl/WalletTest.java
+++ b/src/test/java/io/xpring/xrpl/WalletTest.java
@@ -5,9 +5,13 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+import io.xpring.common.XrplNetwork;
+import io.xpring.xrpl.helpers.XrpTestUtils;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
+
+import java.math.BigInteger;
 
 public class WalletTest {
   @Rule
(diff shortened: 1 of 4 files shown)


src/test/java/io/xpring/xrpl/WalletTest.java
```diff
@@ -197,4 +201,19 @@ public void testVerifyEmptyStringBadSignature() throws XrpException {
...
+    Wallet wallet = XrpTestUtils.randomWalletFromFaucet();
+
+    // WHEN the wallet is examined
+    // THEN it exists
+    assertNotNull(wallet);
+
+    // AND it has a non-zero balance
+    XrpClient xrpClient = new XrpClient("test.xrp.xpring.io:50051", XrplNetwork.TEST);
+    BigInteger balance = xrpClient.getBalance(wallet.getAddress());
+    assertTrue(balance.compareTo(new BigInteger("0")) == 1);
```
keefertaylor, line 217: I think it might be slightly more idiomatic to use `signum` here: https://www.geeksforgeeks.org/biginteger-signum-method-in-java/

src/test/java/io/xpring/xrpl/helpers/XrpTestUtils.java
```diff
@@ -43,4 +55,77 @@
...
+    XrpClient xrpClient = new XrpClient(rippledUrl, XrplNetwork.TEST);
+
+    // Balance prior to asking for more funds
+    BigInteger startingBalance;
+    try {
+      startingBalance = xrpClient.getBalance(address);
+    } catch (Exception exception) {
+      startingBalance = new BigInteger("0");
+    }
+    // Ask the faucet to send funds to the given address
```
keefertaylor, line 79: nit: +1 vertical space above

src/test/java/io/xpring/xrpl/helpers/XrpTestUtils.java
```diff
@@ -43,4 +55,77 @@
...
+    String faucetUrl = "https://faucet.altnet.rippletest.net/accounts";
+    URL url = new URL(faucetUrl);
+    HttpURLConnection con = (HttpURLConnection)url.openConnection();
```
keefertaylor, line 82: optional: Generally prefer to not abbreviate variables unless you really think it's a common abbreviation.

src/test/java/io/xpring/xrpl/helpers/XrpTestUtils.java
```diff
@@ -43,4 +55,77 @@
...
+    con.setRequestMethod("POST");
+    con.setRequestProperty("Content-Type", "application/json");
+    con.setRequestProperty("Accept", "application/json");
+    con.setDoOutput(true);
+    String jsonInputString = String.format("{\"destination\": \"%s\"}", classicAddress);
+
+    try (OutputStream os = con.getOutputStream()) {
```
keefertaylor, line 89: same comment re: abbreviations

src/test/java/io/xpring/xrpl/XrpClientIntegrationTests.java
```diff
@@ -41,14 +43,22 @@
...
+  private static final BigInteger AMOUNT = new BigInteger("1");
 
   /**
-   * Drops of XRP to send.
+   * A Wallet with funds on Testnet.
    */
-  private static final BigInteger AMOUNT = new BigInteger("1");
+  private static Wallet WALLET = null;
+
+  static {
```
theotherian, line 55: I wouldn't initialize this in a static initializer if you can avoid it; I'd do it in a test annotated with `@BeforeClass` instead. When things fail on static initializers in Java it can produce ugly `NoClassDefFoundError` issues that are confusing to track down.

amiecorso: Great, thanks Ian!  I put the wallet initialization in the same place that I noticed `xrpClient` being initialized (in an @Before annotated test) - let me know if there are issues with that I'm not seeing.

PR #307, opened 2020-08-12
Creates `XrpDepositPreauth` Protobuf Wrapper Class

tedkalaw:
This PR implements the class XRPDepositPreauth - a native Typescript class to wrap the DepositPreauth protocol buffer. See also https://xrpl.org/depositpreauth.html

### Context of Change

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
`XrpDepositPreauth` class and conversion unit tests now exist.

## Test Plan

### DIFF
--- a/src/main/java/io/xpring/xrpl/model/XrpDepositPreauth.java
+++ b/src/main/java/io/xpring/xrpl/model/XrpDepositPreauth.java
@@ -51,21 +51,19 @@ public interface XrpDepositPreauth {
    * @see "<https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L159"
    */
   static XrpDepositPreauth from(DepositPreauth depositPreauth, XrplNetwork xrplNetwork) {
-    final String authorize = depositPreauth.getAuthorize().getValue().getAddress();
-    final String unauthorize = depositPreauth.getUnauthorize().getValue().getAddress();
     final boolean isTestNetwork = xrplNetwork == XrplNetwork.TEST || xrplNetwork == XrplNetwork.DEV;
 
     Optional<String> authorizeXAddress = Optional.empty();
     Optional<String> unauthorizeXAddress = Optional.empty();
-    if (authorize != null) {
+    if (depositPreauth.hasAuthorize()) {
       ClassicAddress authorizeClassicAddress = ImmutableClassicAddress.builder()
-          .address(authorize)
+          .address(depositPreauth.getAuthorize().getValue().getAddress())
           .isTest(isTestNetwork)
           .build();
       authorizeXAddress = Optional.of(Utils.encodeXAddress(authorizeClassicAddress));
-    } else if (unauthorize != null) {
+    } else if (depositPreauth.hasUnauthorize()) {
       ClassicAddress unauthorizeClassicAddress = ImmutableClassicAddress.builder()
-          .address(unauthorize)
+          .address(depositPreauth.getUnauthorize().getValue().getAddress())
           .isTest(isTestNetwork)
           .build();
       unauthorizeXAddress = Optional.of(Utils.encodeXAddress(unauthorizeClassicAddress));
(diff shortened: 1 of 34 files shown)


src/main/java/io/xpring/xrpl/model/XrpDepositPreauth.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Optional;
+
+/**
+ * Represents a DepositPreauth transaction on the XRP Ledger.
+ * <p>
+ * A DepositPreauth transaction gives another account pre-approval to deliver payments to the sender of this
+ * transaction.
+ * This is only useful if the sender of this transaction is using (or plans to use) Deposit Authorization.
+ *
+ * Note: authorize and unauthorize are mutually exclusive fields: one but not both should be set.
```
tedkalaw, line 19: I copied https://github.com/xpring-eng/Xpring-JS/pull/493, but...is this sufficient? should we explicitly throw if both are set, or return null, or something?

src/main/java/io/xpring/xrpl/model/XrpDepositPreauth.java
```diff
@@ -0,0 +1,80 @@
...
+   * @see "<https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/proto/org/xrpl/rpc/v1/transaction.proto#L159"
+   */
+  static XrpDepositPreauth from(DepositPreauth depositPreauth, XrplNetwork xrplNetwork) {
+    final boolean isTestNetwork = xrplNetwork == XrplNetwork.TEST || xrplNetwork == XrplNetwork.DEV;
+
+    Optional<String> authorizeXAddress = Optional.empty();
+    Optional<String> unauthorizeXAddress = Optional.empty();
+    if (depositPreauth.hasAuthorize()) {
+      ClassicAddress authorizeClassicAddress = ImmutableClassicAddress.builder()
+          .address(depositPreauth.getAuthorize().getValue().getAddress())
```
tedkalaw, line 63: i copied https://github.com/xpring-eng/Xpring-JS/pull/493 here, but...guessing it makes sense that we don't use a destination tag here?

src/main/java/io/xpring/xrpl/model/XrpDepositPreauth.java
keefertaylor: CC @amiecorso

src/main/java/io/xpring/xrpl/model/XrpDepositPreauth.java
keefertaylor: I believe that we can't authorize / unauthorize at tag fields (and in fact, the tag field doesn't even exist in this transaction type).

src/main/java/io/xpring/xrpl/model/XrpDepositPreauth.java
amiecorso: you know.... Keefer and I kinda discussed this and decided that:  these classes should always be populated by protobufs that are coming over the wire from the ledger (with the exception of our tests where we obviously construct them by hand).  and we are trusting the ledger not to give us bogus data.  There are other places in these classes that have similar but less obvious mutual exclusion or other rules, and we don't really police things there, either.  However, it does seem possible that while the data on XRPL is probably correct, that the gRPC implementation that's sending things back to us may not have been well-vetted at this point.  So I would say - I'm ok with you just leaving things as-is, but it could be worthwhile to add some additional checks at some point if we have any concerns about the quality of the data.  This is actually a task that I thought about potentially asking Mayuka to do (if she's working on SDK at all - not sure) because it would get her familiar with these classes and also give her a concrete task for getting familiar with some of the workings on XRPL.

src/main/java/io/xpring/xrpl/model/XrpDepositPreauth.java
amiecorso: yeah I think this is correct... makes me question whether it makes sense to use an X-address here but I suppose consistency in the SDK is worth it

tedkalaw: i think consistency is good here!

src/main/java/io/xpring/xrpl/model/XrpDepositPreauth.java
tedkalaw: sounds good. i added https://app.asana.com/0/0/1189602713416473/f to track

PR #315, opened 2020-08-18
Adds Utils.isTestNetwork

tedkalaw:
Adds `Utils.IsTestNetwork`, which returns whether or not the passed in `XrplNetwork` is a test network.

### Context of Change
Checks for `network == XrplNetwork.TEST || network == XrplNetwork.DEV` were very common while I was adding transaction types.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
`Utils.isTestNetwork` exists and is used in transaction types

## Test Plan
Run tests + CI. They should pass

### DIFF
--- a/src/main/java/io/xpring/xrpl/Utils.java
+++ b/src/main/java/io/xpring/xrpl/Utils.java
@@ -223,6 +223,6 @@ public class Utils {
    * @return A boolean representing whether or not the {@link XrplNetwork} is a test network.
    */
   public static boolean isTestNetwork(XrplNetwork xrplNetwork) {
-    return xrplNetwork == XrplNetwork.TEST || xrplNetwork == XrplNetwork.DEV;
+    return xrplNetwork != XrplNetwork.MAIN;
   }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/io/xpring/xrpl/Utils.java
```diff
@@ -214,4 +215,14 @@ public static String xrpToDrops(String xrp) throws XrpException {
...
   }
+
+  /**
+   * Check if the XrplNetwork is a test network.
+   *
+   * @param xrplNetwork - The {@link XrplNetwork} to check.
+   * @return A boolean representing whether or not the {@link XrplNetwork} is a test network.
+   */
+  public static boolean isTestNetwork(XrplNetwork xrplNetwork) {
+    return xrplNetwork != XrplNetwork.MAIN;
```
tedkalaw, line 226: i chose to add `isTestNetwork` instead of `isMainnet` because we tend to check whether or not something is a test, not whether or not something "isn't" mainnet, if that makes sense. let me know if you think i should flip it...also if you think the language can be improved.

i also went for a `!= XrplNetwork.MAIN` check to be more defensive

tedkalaw:
per offline discussion, we're tracking the js version of this in a task. would be a good onboarding task :)

PR #316, opened 2020-08-18
Creates `XrpEscrowCreate` Protobuf Wrapper Class

tedkalaw:
This PR implements the class XrpEscrowCreate - a native Java class to wrap the EscrowCreate protocol buffer. See also https://xrpl.org/escrowcreate.html

### Context of Change

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
XrpEscrowCreate class and conversion unit tests now exist.

## Test Plan

### DIFF
--- a/src/main/java/io/xpring/xrpl/model/XrpEscrowCreate.java
+++ b/src/main/java/io/xpring/xrpl/model/XrpEscrowCreate.java
@@ -4,12 +4,29 @@ import org.immutables.value.Value;
 
 import java.util.Optional;
 
+/**
+ * Represents an EscrowCreate transaction on the XRP Ledger.
+ *
+ * An EscrowCreate transaction sequesters XRP until the escrow process either finishes or is canceled.
+ *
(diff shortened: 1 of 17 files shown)


src/main/java/io/xpring/xrpl/model/XrpEscrowCreate.java
```diff
@@ -0,0 +1,138 @@
...
+   * </p>
+   * @return An {@link XrpCurrencyAmount} containing the amount of XRP, in drops, to deduct from the sender's balance
+   *         and escrow.
+   */
+  XrpCurrencyAmount amount();
+
+  /**
+   * (Optional) The time, in seconds since the Ripple Epoch, when this escrow expires.
+   * <p>
+   * This value is immutable; the funds can only be returned the sender after this time.
```
amiecorso, line 40:
```suggestion
   * This value is immutable; the funds can only be returned to the sender after this time.
```

PR #327, opened 2020-08-21
Creates 'XrpSignerListSet' Protobuf Wrapper Objects

tedkalaw:
This PR implements the class XrpSignerListSet, a native Java class to wrap the XrpSignerListSet protocol buffer (See also https://xrpl.org/signerlistset.html)

### Context of Change

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
XrpSignerListSet class and conversion unit tests now exist.

## Test Plan

### DIFF
--- a/src/main/java/io/xpring/xrpl/model/XrpSignerListSet.java
+++ b/src/main/java/io/xpring/xrpl/model/XrpSignerListSet.java
@@ -3,6 +3,11 @@ package io.xpring.xrpl.model;
 import org.immutables.value.Value;
 import org.xrpl.rpc.v1.SignerListSet;
 
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
 /**
  * Represents a {@link SignerListSet} transaction on the XRP Ledger.
  * <p>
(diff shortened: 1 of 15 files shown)


src/main/java/io/xpring/xrpl/model/XrpSignerListSet.java
```diff
@@ -0,0 +1,78 @@
...
+    if (!signerListSet.hasSignerQuorum()) {
+      return null;
+    }
+
+    final Integer signerQuorum = signerListSet.getSignerQuorum().getValue();
+
+    final Optional<List<XrpSignerEntry>> signerEntries = signerListSet.getSignerEntriesList().size() > 0
+        ? Optional.of(signerListSet.getSignerEntriesList().stream()
+          .map(XrpSignerEntry::from)
+          .filter(Objects::nonNull)
```
tedkalaw, line 69: quick q here...

given how protobufs look, it seems like `getSignerEntriesList` is, at worst, an empty `[]`. that seems fine

however, i'm wondering if this `filter` is not actually what we want - you can imagine a world where the array would've been `[null]`, but i've turned it into `[]`, which (based on what this transaction type seems to work) is totally different. should i remove this filter? having an array with a `null` seems strange, but it seems less ambiguous. another possibility is to throw an error if an `XrpSignerEntry` fails to convert

keefertaylor: I think in theory this is possible but in reality this would never happen. I can't imagine a case where you would actually want a null entry in a signers list. 

I think this approach is fine from a data sanitization perspective.

amiecorso: i'm down with that: the list should be clean and `SignerEntry` should probably be throwing errors if there's a problem, which we can do later

PR #344, opened 2020-09-03
port X-address / Classic address encoding and decoding to pure Java

nhartner:

So begins the migration to native Java instead of using Java -> Javascript for XRPL functions. First we begin by rewriting the X-address / Classic address encoding and decoding to pure Java.

### Context of Change

The current architecture uses GraalVM to call XpringCommonJS for XRPL signing, encoding stuffs. This has proven to be slow and does not work on Android. 

This PR rewrites the  X-address / Classic address encoding and decoding to pure Java. Most the code was lifted from ripple-lib-java-unmaintained which is not available as a maven dependency.  

The new class I created from scratch is XAddressCodecs. I also updated the Utils library to call these new classes. And I removed the unused Javascript functions that are now in Java.

Otherwise, all the other classes were copied from ripple-lib-java-unmaintained and cleaned up (mostly formatting and javadoc).

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Unit test coverage is pretty good on the refactored code. Mainly relying on these for test validation.

### DIFF
--- a/src/main/java/io/xpring/codec/addresses/ClassicAddressCodec.java
+++ b/src/main/java/io/xpring/codec/addresses/ClassicAddressCodec.java
@@ -8,13 +8,13 @@ public class ClassicAddressCodec {
   public static final B58 codec = new B58("rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz");
 
   // Otherwise known as `family seed`
-  public static final B58.Version SEED_K256 = new B58.Version(
+  public static final B58.Version SEED_K256_VERSION = new B58.Version(
       33, "seedK256", 16);
 
-  public static final B58.Version SEED_ED25519 = new B58.Version(
+  public static final B58.Version SEED_ED25519_VERSION = new B58.Version(
       "01E14B", "seedEd25519", 16);
 
-  public static final B58.Version ACCOUNT_ID = new B58.Version(
+  public static final B58.Version ACCOUNT_ID_VERSION = new B58.Version(
       0, "accountId", 20);
 
   public static final B58.Version NODE_PUBLIC_KEY = new B58.Version(
@@ -32,24 +32,24 @@ public class ClassicAddressCodec {
   }
 
   public static byte[] decodeSeedToBytes(String seed) {
-    return codec.decodeVersioned(seed, SEED_K256, SEED_ED25519)
+    return codec.decodeVersioned(seed, SEED_K256_VERSION, SEED_ED25519_VERSION)
         .payload;
   }
 
   public static B58.Decoded decodeSeed(String seed) {
-    return codec.decodeVersioned(seed, SEED_K256, SEED_ED25519);
+    return codec.decodeVersioned(seed, SEED_K256_VERSION, SEED_ED25519_VERSION);
   }
 
   public static String encodeSeedK256(byte[] bytes) {
-    return encode(bytes, SEED_K256);
+    return encode(bytes, SEED_K256_VERSION);
   }
 
   public static String encodeAccountID(byte[] bytes) {
-    return encode(bytes, ACCOUNT_ID);
+    return encode(bytes, ACCOUNT_ID_VERSION);
   }
 
   public static byte[] decodeAccountID(String id) {
-    return decode(id, ACCOUNT_ID);
+    return decode(id, ACCOUNT_ID_VERSION);
   }
 
   public static String encodeNodePublic(byte[] bytes) {
@@ -77,6 +77,6 @@ public class ClassicAddressCodec {
   }
 
   public static boolean isValidAccountID(String encoded) {
-    return isValid(encoded, ACCOUNT_ID);
+    return isValid(encoded, ACCOUNT_ID_VERSION);
   }
 }
--- a/src/main/java/io/xpring/codec/addresses/ClassicAddressCodec.java
+++ b/src/main/java/io/xpring/codec/addresses/ClassicAddressCodec.java
@@ -1,0 +1,82 @@
+package io.xpring.codec.addresses;
+
+
+import io.xpring.codec.base58.B58;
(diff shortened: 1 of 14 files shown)


src/main/java/io/xpring/codec/addresses/ClassicAddressCodec.java
```diff
@@ -0,0 +1,82 @@
...
+
+
+import io.xpring.codec.base58.B58;
+import io.xpring.codec.basex.EncodingFormatException;
+
+public class ClassicAddressCodec {
+  public static final B58 codec = new B58("rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz");
+
+  // Otherwise known as `family seed`
+  public static final B58.Version SEED_K256 = new B58.Version(
```
keefertaylor, line 11: Optional nit: you may refer to these as "prefixes" though I understand the library you are using uses "version". 

Either way, `SEED_K256_VERSION` or `SEED_K256_PREFIX` seem like more descriptive variable names without sacrificing too much verbosity.

src/main/java/io/xpring/codec/addresses/XAddressCodec.java
```diff
@@ -0,0 +1,126 @@
...
+   */
+  public static String encode(ClassicAddress classicAddress) {
+    // To compute the X-Address, several hex values are computed and concatenated together, then encoded
+    // using Base58 with a XRPL version
+    try {
+      Prefix prefix = classicAddress.isTest() ? Prefix.TESTNET : Prefix.MAINNET;
+      String addressBytes = base16().encode(ClassicAddressCodec.decodeAccountID(classicAddress.address()));
+      String tagFlag = classicAddress.tag().isPresent() ? TAG_PRESENT_FLAG : "00";
+      String tagBytes = encodeTag(classicAddress.tag());
+      byte [] xaddress = base16().decode(prefix.prefix + addressBytes + tagFlag + tagBytes + RESERVED_SUFFIX);
```
keefertaylor, line 40: optional nit: based on capitalization of  the class I would refer to this as `xAddress`

src/main/java/io/xpring/xrpl/Utils.java
```diff
@@ -1,6 +1,8 @@
 package io.xpring.xrpl;
 
 import com.google.common.base.Preconditions;
+import io.xpring.codec.addresses.ClassicAddressCodec;
```
keefertaylor, line 4: This class was meant to act as a facade, which simply wrapped another class (`JavaScriptUtils`, `CppUtils`). In a very over-engineered world, you could imagine refactoring this class to use some sort of ServiceRegistry/Dependency Injection to pull a native, C++, or JS implementation based on the environment. 

That said, it was definitely a pipe dream to set up a system like described above, and admittedly it may have been over engineered (and the design you are proposing does not preclude that future either). Adding your implementation to a UtilsImpl.java class might preserve that future, but it is up to you whether that is interesting to you.

src/main/java/io/xpring/xrpl/Utils.java
```diff
@@ -86,7 +88,11 @@ public static String encodeXAddress(ClassicAddress classicAddress) {
    */
   @SuppressWarnings("checkstyle:ParameterName")
   public static ClassicAddress decodeXAddress(String xAddress) {
-    return javaScriptUtils.decodeXAddress(xAddress);
+    try {
+      return XAddressCodec.decode(xAddress);
```
keefertaylor, line 92: Just noting from my comment about `xaddress` above that we use `xAddress` here. Whichever way you choose we should try to be consistent.

src/test/java/io/xpring/codec/addresses/XAddressCodecTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.xpring.xrpl.ClassicAddress;
+import org.junit.Test;
+
+public class XAddressCodecTest {
+
+  @Test
+  public void testEncodeMainnetNoTag() {
```
keefertaylor, line 11: optional: Prefer using GIVEN / WHEN / THEN which we've tried to adopt as a standard across Xpring SDK and Xpring's repos in general (https://martinfowler.com/bliki/GivenWhenThen.html). I think it makes tests easier to grok quickly and more readable.

keefertaylor (approved):
Seems reasonable to me. Providing a few optional comments that I think help maintain coding styles and consistency across our codebase, but ultimately I'm not going to bock on them.

src/main/java/io/xpring/codec/addresses/XAddressCodec.java
nhartner: I had `xAddress` originally but then it failed the checkstyle rules so I changed it.

nhartner: OK, I see that you had been suppressing that warning in other classes that use a variable name `xAddress`. I'll follow suit here as well with the suppress warning.

src/main/java/io/xpring/xrpl/Utils.java
nhartner: Yeah, some sort of DI or service registry is going to be needed once there are multiple implementations and to make unit testing easier. For now, I wasn't going to tackle that since it doesn't feel like a big issue.

sappenin (approved):

amiecorso (approved):

sarahkevinking (changes requested):
This is a significant architecture refactor that we need to talk through. I'm going to block these PRs until we discuss as a group on Friday.

intelliot (approved):

sappenin:
Xpring4j is no longer maintained. For Java XRPL support please see https://github.com/xrplf/xrpl4j

PR #346, opened 2020-09-08
Xrp wallet signing rewrite

nhartner:

Rewrote the Wallet signing and validation logic to be pure Java using https://github.com/sublimator/ripple-lib-java 

### Context of Change

This removes all dependency on the JS code for XRP related functions. Only the PayID functions still hit the Javascript code. 
This has 2 benefits:
- much faster
- should work on Android

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Relying on existing test cases and integration tests.


sarahkevinking (changes requested):
This is a significant architecture refactor that we need to talk through. I'm going to block these PRs until we discuss as a group on Friday.

amiecorso (approved):

